PR #88, opened 2026-06-13
Endgame P3-5: reduce memory timeline boundary

codygunton:


codygunton:
Superseded by the P3 closeout: #90 (audit apparatus + obligation map) and #91 (honest reshape to memoryTimelineConstructionEvidence + non-degenerate alignment witness). Those landed the reworked, measurable, documented version of this change.

PR #92, opened 2026-06-14
ci: split Aeneas extraction from proofs gate

codygunton:


codygunton:
Closing this direction. We want the Aeneas extraction tracked and checked by a regeneration/diff gate instead of just splitting the staged extraction out of the proofs job.

Issue #151, opened 2026-06-26
Make RowOutsideDefectRegion row-local: signed-defect Arith witness not trace-projectable (blocked on Arith range-table fidelity)


codygunton:
> **Update (2026-06-29) — re-scoped after #187 (closes #169) landed the Arith range-table fidelity layer.**
>
> Two genuinely separate deliverables were being conflated. #187 fully enables the first; the second (this issue's original Goal) is the hard one. *(An earlier version of this note claimed the `h_sign` discharge shares the provider-row accessor with Deliverable B — that was wrong; it's local, see below.)*
>
> ### Deliverable A — discharge `h_sign_a`/`h_sign_b`. Enabled by #187. LOCAL + mechanical.
> `equiv_MULH_of_table` / `equiv_MULHSU_of_table` already receive `arith_table : ArithMulTableWitness` — which now exposes `arith_table.indexed_ranges : IndexedRangeSpec` (#187) — and `arith_chunk_ranges`. So every ingredient is already in local scope; no provider-row accessor / uniqueness / trace-localization is needed. Recipe:
> 1. Add `mulh_range_pins` (op 181) and `mulhsu_range_pins` (op 179) to `ZiskFv/AirsClean/ArithTableProjections.lean`, mirroring the existing **`mul_range_pins`** (op 180, line 313) — same `fin_cases`-over-ROM proof. **Verified against `zisk/state-machines/arith/src/arith_table_data.rs`:** the ROM rows give, for MULH (181), `na=0 ⇒ range_ab∈{4,5}` (POS), `na=1 ⇒ range_ab∈{7,8}` (NEG), with `nb` via `range_ab+17 ∈ {21,24}` POS / `{22,25}` NEG; for MULHSU (179), `na=0 ⇒ range_ab=3` (POS), `na=1 ⇒ range_ab=6` (NEG), `nb` pinned 0. (`ArithRange{Pos,Neg}Id` membership of all these is already proved in `RangeTables`.)
[...]

PR #187, opened 2026-06-29
Add arith indexed range table fidelity

### LINKED ISSUE #169: Arith range-table fidelity for row-local defect witnesses
## Problem

#151 is blocked on Arith range/table fidelity work that currently has no live issue node. #151 says the signed-defect witness path needs the indexed sign-range lookup to be balance-derivable, and that the related closed #114 / PR #121 did not cover this specific path.

## Scope

- Compose/extract the Arith range/table AIRs needed for the signed-defect path instead of leaving the relevant range-table facts collapsed to `FULL`.
- Make the indexed sign-range lookup balance-derivable for the `na = MSB` / signed-defect witness path.
- Provide the extraction/fidelity layer needed before the Lean-side provider-row accessor and uniqueness proof in #151 can land.

## Acceptance

- The Arith range/table facts needed by #151 are present in the live extracted/composed AIR model.
- #151 can start its Lean-side row-local defect proof without treating this range-table fidelity step as an external blocker.

Blocks #151.

codygunton:
Closes #169.

## Summary
- add the Clean static indexed `arithRangeTable` model and POS/NEG projection lemmas
- thread the eight upstream indexed Arith range lookups through ArithMul/ArithDiv specs, constraints, bridge projections, and compliance witnesses
- add row-local `na`/`nb` MSB derivation lemmas for signed 64-bit and W paths
- update trust notes to distinguish the new in-model indexed range evidence from public `h_sign_*` binders that #151 still needs to wire

## Verification
- `lake build ZiskFv.AirsClean.ArithTableProjections ZiskFv.Compliance.SharedBundles ZiskFv.Compliance.Wrappers.MulH ZiskFv.Compliance.Wrappers.MulHSU ZiskFv.Compliance.TraceLevelExport.StepStrongLoadMext`
- `git diff --check origin/main...HEAD`
- refined touched-file Lean placeholder scan for actual `sorry`/`admit`/`axiom`/`native_decide` additions
- `lake build`


codygunton (commented):
## Review — sound and faithful; correctly scoped as an enabling step, not a discharge

Post-merge audit focused on the two things only a human can check here: **upstream fidelity of the new hand-maintained model** (the Arith range tables are not extraction-checked) and **anti-laundering scope** (does the trust surface move honestly?). Both hold up. CI on the merge is green (`lake build (full FV check): success`, `trust-gate: success`).

### Fidelity — verified line-by-line against ZisK source ✅

The new `RangeTables.arithRangeTable` claims to be an "exact static model of upstream `arith_range_table.pil`." I checked it against `zisk/state-machines/arith/pil/arith_range_table.pil` and `arith.pil`, and it is exact:

- **The 8 indexed lookups** added to `mainWithArithTable` match `arith.pil:299-306` verbatim — same chunks (`a1,b1,c1,d1,a3,b3,c3,d3`) and same offsets. The offset map (`a3=+0, a1=+26, b3=+17, b1=+9`, c/d via `range_cd`) matches the `const expr range_*` defs at `arith.pil:289-297` and the col-compression table at `arith_range_table.pil:50-53`.
- **Table dimensions:** `length = 2^22 = 4194304` ✓; `arithRangeChunkRows = 25·65536 + 18·32768 = 2228224` ✓; carry region = 1966080 rows ✓ (sums to 2^22).
- **`arithRangeHalfBlockId`** reproduces the `RANGE_ID` fixed column (`arith_range_table.pil:61-64`) exactly across all 68 half-blocks — including the FULL-range pairing (blocks `2k,2k+1 → k`-th FULL id over `{0..2,9..17,20,23,26..36}`) and the POS/NEG singleton blocks (50→3 … 58→39, 59→6 … 67→42).
- **`ArithRangePosId = {3,4,5,18,21,24,37,38,39}`** and **`ArithRangeNegId = {6,7,8,19,22,25,40,41,42}`** match the PIL's POS/NEG id sets exactly.
- **The bound lemmas are the right reading of `RANGE_VALUES`** (`pil:66-69`): POS ⊂ `[0x0000..0x7FFF]` ⇒ `< 32768` (`arithRangeTable_pos_bound_of_spec`); NEG ⊂ `[0x8000..0xFFFF]` ⇒ `[32768, 65536)` (`_neg_bound_of_spec`). This is exactly the sign mechanism: a POS/NEG id on the **high** chunk (`a[3]` for 64-bit, `a[1]` at offset 26 for W-mode bit-31) pins MSB. `packed4`/`packed2` and the 64- vs 32-bit split are correctly matched to which chunk carries the sign.
- **Carry encoding** `arithRangeCarryValue` correctly models `[-0xEFFFF..0xF0000]` in field form.

This is careful, genuinely faithful work — the kind of fidelity claim that's easy to get subtly wrong and didn't.

### Soundness / non-vacuity ✅

- `IndexedRangeSpec` is *derived*, not assumed: `ArithMul/Circuit.lean` produces it from the FormalCircuit's own `h_holds` (the 8 lookups), and the unsigned Div path inherits it via `arithDiv_fullSpec_of_arithMul_fullSpec`. So strengthening `FullSpec` with the new conjunct does not introduce vacuity — real rows witness it because the real circuit enforces these lookups (`arith.pil:299-306`).
- The `na/nb_eq_msb{64,32}_of_{pos,neg}_indexed` lemmas are correct *conditional* helpers: they take both the case (`h_na : na = 0/1`) and the matching `h_id`, and derive `na = MSB` from the range bound. Not circular, not yet a discharge — the missing piece is the `na ↔ range_ab`-id linkage from the ArithTable ROM, which is explicitly #151's job.
- `IndexedRangeSpec` is marked `@[reducible]`, so the V2 `whnfR` binder gate sees through it (CLAUDE.md trap #5 / definitional-aliasing) — correct call.

### Anti-laundering scope ✅

This is the judgment that matters most for a trust-ledger PR, so to be explicit: **this PR does not reduce the canonical trust surface, and correctly does not claim to.** `equiv_MULH`/`equiv_MULHSU` still carry `h_sign_a`/`h_sign_b`; the per-theorem `collectAxioms` closure is unchanged (0 `ZiskFv.*` axioms); no new axiom/sorry/`Valid_*` over-constraint. By the strict "net-zero PRs should be rescoped" rule that would be a flag — but #169's acceptance criteria are *enabling* ("facts present in-model so #151 can start"), and this PR meets them: it moves `na = MSB` from "unprovable in-model" to "row-locally derivable from a faithful in-model lookup." The `defects.md` / `trusted-base.md` reframing is honest about exactly this — it drops the old "unprovable in-model / collapsed to FULL" language and states the public binders remain until #151 wires provider-row accessors. No overclaim.

One framing note worth recording: the 8 lookups are *hand-added* to the Clean `mainWithArithTable`, consistent with how that model already mirrors `arith.pil` (ArithTable lookup, chunk ranges, carries). So this extends the existing hand-maintained-fidelity methodology rather than deriving from extracted balance — no new trust *kind*, and the mirroring is now source-verified above.

### Minor, non-blocking

- `arithRangeTable`'s carry region (id 100, `arithRangeCarryValue`) is modeled for table-completeness but isn't consumed by any current path — carries still flow through the separate `signedCarryRangeTable`, and the bound lemmas discharge the carry case via `arithRange{Pos,Neg}Id_not_100`. Fine (it makes the model a complete image of the single upstream table); just noting it's currently inert.
- The two `interval_cases block` proofs (`_lt_32768_of_pos`, `_neg_bounds_of_neg`) need `maxHeartbeats 1000000` for the 68-way split — acceptable for a one-time faithfulness proof, but a build-cost hotspot to be aware of.

Net: approve (would, if it weren't my own PR). The load-bearing fidelity is verified against source, the soundness chain is non-vacuous, and the trust accounting is honest. #151 is unblocked.

PR #189, opened 2026-06-29
Derive initial trace placement facts

codygunton:
Part 1 of the review-sized #141 stack.

This first chunk adds the foundational trace placement derivations used by the later removals:

- model/project Main address-placement constraints from the accepted trace;
- expose row-level address placement through `RomDecodeBinding`;
- remove the first ADD/ADDI/SUB destination-placement assumptions from public `Inputs_*` records;
- derive those destination facts from `AddressSpec` plus committed-program decode.

Review stack after consolidation:

1. #189 initial trace placement facts
2. #196 ALU and M-extension rd placement
3. #199 U-type/jump/load/store placement
4. #204 signed-M/control/source-lane cleanup
5. #208 immediate and offset decode bridges

Size: about +734/-134 lines.

Verification already run on this head before opening: focused Lean gates, `lake build`, `trust/scripts/check-all.sh`, and `trust/scripts/check-all-semantic.sh`.

Issue #141, opened 2026-06-24
root_soundness link assumes each op's effect *placement* (dest reg / store address / nextPC) instead of deriving it from the Main AIR


codygunton:
## Summary

`root_soundness` no longer has the original monolithic `rowData` / `OpEnvelope` surface, but the remaining per-row input bundle still mixes genuine cross-world agreement with facts that should be derived from the accepted ZisK trace.

The current residual for this issue is **effect placement**: destination register, load/store target address, and the remaining control/range facts that decide where the already-proved value lands.

## Current status (2026-06-29)

Several parts of the original issue have already been reduced:

- #143 split the old row-data surface into `ZiskStep` / `RowDecode` / `InputsAgree`.
- #147 parameterized `AcceptedZiskTrace` / `SailTrace` by the same instruction count.
- #163, closing #100, removed the old caller-supplied `h_nextPC_matches` obligation by deriving the next-PC equality for all 63 RV64IM ops from the accepted trace / Main transition facts.
- #170 changed the `root_soundness` decode premise from caller-supplied row decode columns to `ProgramDecode`; `RowDecode` is now derived with `rowDecode_of_programDecode` through the committed program / ROM lookup path.
- #100 and #101 are closed and should no longer be treated as blockers for this issue.

The live headline signature is now:

```lean
root_soundness
  (numInstructions : Nat)
  (ziskTrace : AcceptedZiskTrace numInstructions)
  (sailTrace : SailTrace numInstructions)
  (ziskStep : forall i, ZiskStep ziskTrace i)
  (programDecodes : forall i, ProgramDecode ziskTrace i (ziskStep i))
  (inputsAgree : forall i, InputsAgree ziskTrace sailTrace i (ziskStep i))
  (hAvoidKnownBugs : forall i,
    RowOutsideDefectRegion ziskTrace sailTrace i (ziskStep i) (inputsAgree i))
  : forall i, StepSound ziskTrace sailTrace i (ziskStep i)
```

## Remaining gap

`InputsAgree` still contains placement/control facts that should eventually be derived rather than supplied:

- destination register pins such as `h_rd_idx` across ALU/shift/I/W, LUI/AUIPC, and jump/link rows;
- store target address pins such as `h_addr2` for `SB`/`SH`/`SW`/`SD`;
- load address/destination pins such as `h_addr1`, `h_addr2_zero_iff`, and `h_addr2_idx` for load rows;
[...]

PR #190, opened 2026-06-29
Derive logical compare rd placement

codygunton:
Stacked on #189; part of #141 and does not close it.

## Summary
- derive AND/OR/XOR/SLT/SLTU writeback destination placement from committed-program store selector/offset facts plus AddressSpec
- remove the corresponding input-side h_rd_idx fields from RowDataAluShift
- add shared mainWritebackDestinationFacts_of_program for the op-agnostic Main row and use it in ROM decode builders

## Verification
- lake build
- trust/scripts/check-all.sh
- trust/scripts/check-all-semantic.sh


codygunton:
Superseded by review-sized #141 stack chunk #196.

Issue #221, opened 2026-07-01
root_soundness instantiation: load with a real multi-row memory prefix (non-degenerate memory, #74)


codygunton:
## Goal

The non-degenerate-memory OPEN criterion of #74: instantiate `root_soundness` on
a trace whose Mem table carries a **genuine multi-row memory timeline**, so the
load-arm memory obligation is discharged at a concrete, non-degenerate instance
(not the empty-prefix witness in `trust/consistency/global_theorem_instantiation_ld.lean`).

## Scope

1. A tiny store-then-load program (e.g. `SD` then `LD` to the same address) so the
   Mem provider table has a real timeline prefix (a store row, then the load's
   read).
2. Extend the #220 instantiation with a `ZiskStep.ld` (and `ZiskStep.sd`) arm:
   build `Claim_ld` / `Decode_ld` / `Inputs_ld`, discharge the load's
   `MemoryTimelineEvidence` / `RowTraceCoherence` obligation at the concrete
   prefix. Starting points: `trust/consistency/memory_prefix_alignment_witness.lean`
   and `load_byte_agreement_witness.lean`.
3. Forward `BalancedChannels` with the multi-row Mem timeline (MemBus now carries
   store + load entries that must cancel against Main's accesses).

## Dependencies

- #76 (memory-timeline reduction to `RowTraceCoherence`) — landed.
- #115 (eliminate the `RowTraceCoherence` floor) — interacts: this concrete
  instance must *supply* the `RowTraceCoherence` witness for its prefix; if #115
  reduces that premise, this instance simplifies accordingly.

## Acceptance

- A checked-in `root_soundness` instantiation on a real multi-row memory-prefix
  trace fragment (store + load), memory obligation discharged concretely.
- Header documents the program + regen recipe.
- No new axioms, no `sorry`; V1+V2 gates green.

Part of #74 (blocks it). Depends on #220; gated on #76 / #115. Satisfies the OPEN
"fully non-degenerate multi-row memory prefix" acceptance criterion of #74.

codygunton:
Implementation is complete in the two-PR stack:

- Phase 1 Mem provider: #290 (`a0ead1ca`)
- Phase 2 accepted trace + root theorem: #291 (`f8c04ef0`), stacked on #290

[...]

Issue #185, opened 2026-06-29
Name the boot/cross-segment memory seed as the single documented cross-row trust premise of root_soundness


codygunton:
## Goal

Give the cross-row memory/PC residual a **single, named, defensible** home so the
`root_soundness` interface is legible even before the residual is fully derived
away. Splits the memory-coherence work into:

- **this issue (MVP, legibility):** define one explicitly-documented
  boot / cross-segment **seed** premise (initial memory + PC state at segment
  entry) and route the load/store coherence residuals through it, so a reader sees
  *one* named cross-row trust premise rather than scattered per-op fields
  (`RowTraceCoherence`, `h_m1..h_m7`).
- **#115 / #119 (beyond-MVP, derivation):** drive that seed-relative residual to
  zero by deriving the Sail memory sequence as the replay of the trace's writes.

## Why split it this way

The MVP bar is "every premise is derived OR named as defensible trust." A boot /
cross-segment seed is genuinely irreducible at the single-segment level (the
segment doesn't contain its own initial state), so *naming* it is legitimate and
sufficient for legibility. Full replay derivation (#115/#119) then strengthens
beyond MVP.

## Scope

1. Introduce a single `BootSegmentSeed` (or similarly named) premise capturing
   initial mem/PC, documented as cross-segment trust.
2. Express `RowTraceCoherence` (loads, #115) and `h_m1..h_m7` (stores, #119) as
   consequences *relative to* that seed, collapsing the per-op memory fields into
   the one named premise where possible.
3. Document the seed in the `root_soundness` premise audit (see the premise-audit
   issue) as part of the irreducible-trust core.

## Acceptance

- `root_soundness` carries at most one named, documented cross-row memory/PC seed
  premise; per-op memory coherence fields are derived from it or explicitly
  reduced to it.
- No new `ZiskFv.*` project axioms.

Blocks #61. Relates to #115, #119, #141.

codygunton:
**Plain-language summary**

*What the project proves:* for each RISC-V instruction, ZisK's circuit is shown to do the same thing as the official RISC-V spec. The top-level theorem (`root_soundness`) chains these together for a whole run of instructions.

[...]

Issue #115, opened 2026-06-20
Eliminate the RowTraceCoherence memory trace-coherence floor — follow-on to #76


codygunton:
> [!WARNING]
> ## ⚠️ SCOPE AMENDMENT — owner-authorized, 2026-07-08
>
> This issue's acceptance criteria are **amended** as follows. The original text is preserved
> below, unedited, for the audit trail. Reviewers of the closing PR should check the amended
> criteria, not the original ones.
>
> **What stays exactly as originally written:** the opaque semantic premise
> (`readSound` / `RowTraceCoherence`) is eliminated — memory-bus read-soundness is **derived**,
> not caller-supplied; no new `ZiskFv.*` project axioms; the non-degenerate memory witnesses stay
> satisfiable; `memInit`/`boot` remains the one deliberately irreducible single-segment premise.
>
> **Two amendments, with reasoning:**
>
> 1. **Direct-Mem scope.** Read-soundness is derived (store-inclusively) for accesses served by
>    the mutable Mem table. **MemAlign-routed accesses** (unaligned / sub-doubleword) retain
>    *named structural residues* rather than a derivation. Reason (forced): MemAlignRom is not
>    extracted as a first-class Lean table, and MemAlign provider rows do not appear in the Mem
>    replay row list, so their read-soundness needs both extraction work and a new
>    through-MemAlign timeline argument. That tail is tracked in the follow-on issue
>    **"MemAlign-family read-soundness discharge"** (gated on the #108 extraction axis) — see the
>    linked issue; it enumerates the exact residues.
> 2. **Certificate-class premises.** In addition to the boot/cross-segment premise, the derivation
>    consumes named, PIL-cited certificate fields on `AcceptedZiskTrace` (the `mem_replay_*`
>    family: generated Mem cross-row constraint facts, row/segment range facts, sidecar source
>    columns; plus, if their timeboxed derivation fails, two step-counter no-collision facts).
>    Reason (forced): the composed Clean Mem component cannot carry its cross-row constraints
>    until #226 is fixed upstream (a component cannot have both channels and a transition), so
>    full derivation from `constraints_hold` alone is impossible today. These fields are the same
>    verifier-checked class as the pre-existing `transitions_hold` / `main_height` /
>    `segment_l1_fixed` certificates. Decision recorded in #237.
>
> **Amended closing criteria:** store-inclusive direct-Mem read-soundness derivation landed; every
[...]

Issue #74, opened 2026-06-12
Instantiate the global theorem on a real ZisK execution trace (end-to-end integration witness)


codygunton:
## What this is, in plain terms

The current trace-level result is `ZiskFv.Compliance.root_soundness`: from an `AcceptedZiskTrace`, a `SailTrace`, per-row `ZiskStep`, per-row `ProgramDecode`, per-row `InputsAgree`, and per-row known-defect exclusions, every row satisfies the strengthened channel-balance conclusion. The old `zisk_riscv_compliant_program_bus` theorem still exists underneath as the internal `OpEnvelope` theorem.

A theorem of this shape still has a standing non-vacuity question: can the quantified-over objects actually be produced from a real ZisK execution? This issue tracks checked concrete instantiations that answer that question with actual trace fragments rather than only per-component witnesses.

## Status (2026-06-29)

- `trust/consistency/global_theorem_instantiation_add.lean` instantiates `zisk_riscv_compliant_program_bus` on a concrete ADD envelope.
- `trust/consistency/global_theorem_instantiation_ld.lean` instantiates `zisk_riscv_compliant_program_bus` on a concrete LD envelope. It is the empty-prefix (`priorRows = []`) load-memory witness, rebuilt to the `LoadMemoryTimelineCoherenceEvidence` / `RowTraceCoherence` shape.
- `trust/consistency/memory_prefix_alignment_witness.lean` covers a concrete single-store prefix for the memory replay side, but it is not yet a fully non-degenerate multi-row global theorem instantiation.
- The next close-path target is a current-surface `root_soundness` instantiation for a concrete trace fragment, likely starting with ADD, with explicit concrete `ProgramDecode`, `InputsAgree`, and `RowOutsideDefectRegion` witnesses.
- The stricter decode-driven/no-caller-supplied-placement witness remains valuable, but it belongs under #61 and its child work (#141/#151/#115/#119/etc.). It should not block this issue's current-surface witness.

## Technical starting point

Target artifact: a checked-in file following the `trust/consistency/global_theorem_instantiation_*.lean` pattern, discovered by the semantic gate, containing:

1. A concrete trace fragment from a real ZisK execution of a tiny RV64IM program.
2. Lean definitions building the corresponding `AcceptedZiskTrace`, `SailTrace`, per-row `ZiskStep`, `ProgramDecode`, `InputsAgree`, and any internal `OpEnvelope` values needed by the proof.
[...]

PR #227, opened 2026-07-06
#115: make the BootSegmentMemorySeed memory premise concrete (constructibility for #221/#74)

codygunton:
## Context

#185 unified the ten per-op memory residuals into one `BootSegmentMemorySeed` `root_soundness` premise, but stated its memory-evolution assumption as an **opaque** free cursor `stateAt` + a whole-sequence `coherence : RowTraceCoherence stateAt [] rows`. #115's remaining acceptance is that the memory premise be stated as "named boot/cross-segment premises, **not** an opaque whole-Sail-state equality."

This PR restates the seed **concretely**.

## What this does

- **`ZiskFv/ZiskCircuit/MemTimeline/Spike.lean`** — adds three kernel-clean lemmas: `rowTraceCoherence_of_uniformReplayMem` (which proves `coherence ⟺ step + boot`), `exec_order_fold_fin`, and `exists_flatMap_range_split_of_singleton`.
- **`BootSegmentMemorySeed.lean`** — the struct's `stateAt`/`h_seed`/`coherence` become concrete `memInit`/`rowsOf`/`boot`/`step`/`readSound`; `MemoryOpPlacement` becomes a structural row tie (`rowsOf i = [busLd..e1]` for loads, `[busSt..e2]` + preserved bytes for `sb`/`sh`/`sw`). `memEvidence_of_bootSeed` now **derives** each op's residual via the execution-order fold (`exec_order_fold_fin` for the state pin + `loadEvidence_of_loadMemReplay`/`storeEvidence_of_loadMemReplay`). The binder **type name is unchanged**, so `root_soundness`'s signature and the export-binder baseline are unaffected.
- **`BootSegmentMemorySeedWitness.lean`** — the non-vacuity witness rebuilt via the concrete reduction.
- **`trust/consistency/root_soundness_instantiation_degenerate.lean`** — supplies an explicit n=0 seed. (This witness was underapplied on `main` — it relied on `nomatch` building the old seed and omitted the `hAvoidKnownBugs` argument; both are fixed here.)
- **`trust/trusted-base.md`** — ledger updated.

## Honest framing (please read)

This is **net-zero on trust strength — NOT a reduction.** `coherence` and `step + boot` are provably equivalent (`rowTraceCoherence_of_uniformReplayMem`), and `readSound` is the same read-soundness the seed already carried. The value is **constructibility**: the memory premise is now concrete, with no opaque `stateAt` existential and no whole-sequence `RowTraceCoherence` on the assumed surface. That is the seam #221 (non-degenerate load instantiation) explicitly asks for — "*if #115 reduces that premise, this instance simplifies*" — feeding #74's headline non-vacuity, and it satisfies #115 acceptance #1.

**What this does NOT do:** #115 acceptance #2 ("derived from accepted trace constraints / channel balance") is not met for the read-soundness half — `readSound` (= `prefixReadSound`) is the out-of-scope ExtF memory-bus **permutation** trust and is not derivable from `constraints_hold`/`channels_balanced`. It remains a named premise of that existing trust class; its full derivation is a separate epic.

## Verification

- `lake build` — green.
- V1 `trust/scripts/check-all.sh` — 16/16.
- V2 `trust/scripts/check-all-semantic.sh` — 15/15. `root_soundness` closure baseline **unchanged**; **0 new project axioms** (closes over kernel + the baselined Sail-spec FFI axioms only).
- The non-degenerate memory witness remains satisfiable; the degenerate `root_soundness` instantiation compiles.

Refs #115, #221, #74.

### DIFF
--- a/ZiskFv/Compliance/TraceLevelExport/BootSegmentMemorySeed.lean
+++ b/ZiskFv/Compliance/TraceLevelExport/BootSegmentMemorySeed.lean
@@ -3,159 +3,216 @@ import ZiskFv.Compliance.TraceLevelExport.Dispatcher
 /-!
 # The boot / cross-segment memory seed (`BootSegmentMemorySeed`)
 
-`root_soundness`'s loads and stores each consume a memory-coherence fact — "the
-value in memory at this address is what the circuit claims" — that no single
-instruction can supply, because it depends on the whole history of earlier
-writes.  Historically the ten memory ops (seven loads + `sb`/`sh`/`sw`) each
-carried their *own* existential copy of that story as an `Inputs_<op>` field
-(`h_memory_timeline` / `h_store_rmw`), so nothing forced the ten copies to agree.
+`root_soundness`'s loads and stores each consume a memory-coherence fact — "the value in memory at
+this address is what the circuit claims" — that no single instruction can supply, because it depends
+on the whole history of earlier writes.  #185 unified the ten former per-op residuals (seven loads +
+`sb`/`sh`/`sw`) into ONE named premise, the `BootSegmentMemorySeed`.
 
-This module replaces those ten scattered residuals with ONE named premise, the
-`BootSegmentMemorySeed`: the segment's initial memory state plus the single
-consistent per-row memory-evolution chain (`RowTraceCoherence` over the *whole*
-consumed memory-bus row sequence).  `memEvidence_of_bootSeed` then *derives* each
-op's residual from that shared seed, restricting the whole-sequence coherence to
-the op's prefix via `rowTraceCoherence_of_append`.
+**Issue #115 (this form).** #185's seed stated its memory-evolution assumption as an *opaque* free
+cursor function `stateAt : List rows → SailState` plus a whole-sequence
+`coherence : RowTraceCoherence stateAt [] rows`.  This form states it *concretely* instead:
 
-This is the MVP legibility step of issue #185: the seed is legitimately *assumed*
-(a segment does not contain its own starting state — it is carried in from the
-previous segment / boot).  Driving the seed to zero is the follow-on work in #115
-/ #119.  Only the memory field of each cursor state is constrained; PC /
-registers are pinned only incidentally through the initial-state snapshot
-(per-step next-PC is discharged separately, #163), which is why this is a
-*memory* seed rather than a "memory+PC" one. -/
+* `memInit` + `boot` — the segment's boot / cross-segment seed memory (irreducible at the single-
+  segment level: a segment does not contain its own starting state, it is carried in from boot);
+* `step` — the per-step **execution-successor**: each Sail step's memory is the replay of that step's
+  memory-bus rows onto the previous step's memory (cross-row memory coherence);
+* `readSound` — memory-bus **read-soundness** over the whole execution-order row list.  This is the
+  out-of-scope ExtF memory-bus **permutation** trust (see `trust/trusted-base.md`;
+  `Airs/MemoryBus/MemBridge.lean`), carried as a named premise of that existing class;
+* `placement` — the *structural* tie pinning `rowsOf i` to each op's real memory-bus row (a load's
+  read `busLd .. .e1`, a narrow store's write `busSt .. .e2` plus its preserved-byte prefix fact).
+
+`memEvidence_of_bootSeed` derives each op's `MemoryOpEvidenceFor` residual by the execution-order fold
+(`Spike.exec_order_fold_fin` gives the per-op state pin from `boot` + `step`;
+`Spike.exists_flatMap_range_split_of_singleton` locates the op's row; `loadEvidence_of_loadMemReplay` /
+`storeEvidence_of_loadMemReplay` build the evidence).
+
+This is **net-zero on trust strength** versus #185's opaque form — `coherence ⟺ step + boot`
+(`Spike.rowTraceCoherence_of_uniformReplayMem`), and `readSound` is the same read-soundness `facts`
+already carried.  Its value is *constructibility*: the memory premise is now concrete, with NO opaque
+`stateAt` existential and NO whole-sequence `RowTraceCoherence` on the assumed surface (#115 acceptance
+"named premises, not an opaque whole-Sail-state equality"), which is the seam the non-degenerate load
+instantiation (#221 → #74) needs.  It is a named external-trust premise (same class as channel-balance),
+NOT an axiom and NOT a defect; the read-soundness half remains the memory-bus permutation trust, whose
+full derivation from `constraints_hold`/`channels_balanced` is a separate epic. -/
 
 namespace ZiskFv.Compliance
 
 open Interaction
+open ZiskFv.ZiskCircuit.MemTrace
+open ZiskFv.ZiskCircuit.MemTimeline.Spike
 
 variable {numInstructions : Nat}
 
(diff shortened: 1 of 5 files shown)


codygunton:
## Review

**Verdict: approve, with two wording fixes + one process caveat.** The refactor is correct, the soundness claims hold up, and — importantly — this PR also repairs a `main` that has been red for five days.

### This PR fixes a red `main` (please lead with this)

The PR frames itself as a constructibility nicety; it's more. Tracing the degenerate-witness change through CI:

- `main`'s last **completed** `proofs` run (full `nix run .#test` / V2 gate) is a **failure** — commit for #185 (PR #224), 2026-07-01:
  ```
  ::: 13/15 root_soundness instantiation witnesses :::
  trust/consistency/root_soundness_instantiation_degenerate.lean:147:47: error: Missing cases:
  trust/consistency/root_soundness_instantiation_degenerate.lean:147:2: error: Type mismatch
  ❌ 7/8 trust gate (V2 semantic) failed
  ```
- Cause: #223 added the witness as `root_soundness 0 trace sail step nofun nofun nofun` (correct when `root_soundness` had 3 trailing params). #185 then inserted `bootSeed : BootSegmentMemorySeed` before `hAvoidKnownBugs` without updating the witness → the third `nofun` lands on the `bootSeed` **structure** position (`nofun` can't inhabit a structure → "Missing cases") and the call is underapplied (missing `hAvoidKnownBugs`). The file is unchanged on `main` since #223, so HEAD is still broken.

So the "underapplied on main" claim is **verified**, and this PR's `... nofun nofun seed nofun` + explicit degenerate `seed` is what turns `main` green again.

**Process caveat (not this PR's fault):** `proofs.yml` triggers only on `push` to `main`; `trust-gate.yml` runs on PRs but only V1 (syntactic, ~25s). So **PRs are never V2-gated before merge** — `main` only goes red *after* a bad merge, which is how #185 slipped through. This PR is also unverified by CI. Please run `nix run .#test` locally before merging, and consider adding a `pull_request` trigger to `proofs.yml` (separate follow-up).

### Soundness / anti-laundering — clean

- **Non-vacuity preserved.** `storeRmwPreservedBytes_nonvacuous` still requires **non-empty** seed memory (false over `{}`), and `witnessStore_evidence` builds a genuine `StoreRmwMemoryCoherenceEvidence` off it. The #185 store-side "empty-mem floor is false" crux survives.
- **`readSound` honestly still trust.** It flows through as `MemoryBusRowsPrefixReadSound` / `facts.prefixReadSound` (the same ExtF permutation trust), not trivialized. The re-packaging inside `loadEvidence_of_loadMemReplay` reuses the input's read content. The PR is explicit that #115 acceptance #2 is not met here — accurate.
- **Closure/axiom footprint.** `root_soundness`'s signature is byte-identical to `main`; `memEvidence_of_bootSeed` returns the same `MemoryOpEvidenceFor` type; the three new `Spike` lemmas carry `#print axioms` guards. "0 new project axioms / baseline unchanged" is structurally consistent (subject to the local build).

### Two wording fixes (the substantive ask)

1. **`coherence ⟺ step + boot` is overstated.** `rowTraceCoherence_of_uniformReplayMem` proves only the `⟸` direction (the uniform-replay cursor *satisfies* coherence, i.e. `step+boot` reconstructs the evidence the old form supplied). There is no `⟹` lemma, and the two sides quantify over different objects (a free `stateAt` cursor vs. the real `binding`). If anything the new premise is marginally **stronger** — `step` pins `binding.mem` at *every* index where the old form tied it only at memory-op indices — which real traces satisfy trivially (no vacuity), but that's "net-zero-to-slightly-stronger," not a biconditional. Per the repo's anti-laundering "precise claims" discipline, please soften the module docstring and the `trusted-base.md` ledger line to state the reconstruction direction actually mechanized, not `⟺`.

2. **`witnessStore_nondegenerate` is now decoupled from the evidence.** The deleted `witnessStore_rowTraceCoherence` wired the mutating `witnessStoreStateAt` cursor into the evidence; the evidence now uses the uniform-replay cursor (regs/PC frozen, only `.mem` evolves), so `witnessStore_nondegenerate` + `witnessStoreStateAt` are a standalone side-claim no longer describing the actual inhabitant. Not a soundness gap (the preserved-bytes non-vacuity is still wired in), but the "not the frozen-state floor" narrative now rests on the wrong object. Either re-tie the non-degeneracy claim to the uniform-replay cursor's `.mem` evolution, or drop the vestigial theorem.

### Minor

- `MemoryOpPlacement` imposing `True` on non-memory ops relies on `step`'s satisfiability to force `rowsOf j = []` there — a constructibility obligation on the witness-builder, correctly documented. Just noting it's the one tie that's "by satisfiability" rather than explicit.

_Note: I could not independently run the full `lake build` (the shared `build/` is stale for this base), so the compile/gate claims are reasoning-verified from source + the author's reported local run, not re-executed._

codygunton:
Thanks — both wording fixes addressed in `a460d43c`, and the red-`main` framing is well taken.

1. **`coherence ⟺ step + boot` softened.** You're right that `rowTraceCoherence_of_uniformReplayMem` mechanizes only the reconstruction direction (`step`+`boot` ⟹ the uniform-replay cursor *satisfies* `RowTraceCoherence`; no converse), and that `step` pins `binding.mem` at every index where the opaque chain tied it only at memory-op indices — so it's net-zero-to-marginally-*stronger*, satisfied trivially by real traces. The Spike + `BootSegmentMemorySeed` docstrings and the three `trusted-base.md` lines now say "constructibility restatement, not a reduction" and state the direction actually proven.

2. **`witnessStore_nondegenerate` dropped.** Correct that it described `witnessStoreStateAt` (the old mutating cursor), which the concrete evidence no longer uses (uniform-replay cursor: regs/PC frozen, only `.mem` evolves) — so it described the wrong object. Removed both it and `witnessStoreStateAt`. The load-bearing non-vacuity crux (`storeRmwPreservedBytes_nonvacuous` off **non-empty** memory, wired through `witnessStore_evidence`) is untouched; the "not a frozen whole-state floor" property is a property of the `LoadMemoryTimelineCoherenceEvidence` type (only `.mem` constrained), documented in the #76 floor section.

3. **Process caveat.** Running the full `nix run .#test` locally before merge, per your note that PRs are only V1-gated. Will confirm the result here.

codygunton:
Ran the full suite locally (`nix run .#test`), and it caught a real issue worth flagging: my review-fix docstring wrapped so a line began with `opaque form`, which the **V1 locality check greps as a trust-leak keyword** → V1 failed. (Notably V1 is the *only* gate CI runs on PRs, so this would have been the sole PR-visible signal.) Reworded to `free-cursor`/`free` phrasing in `411677a1`; no line-initial banned keyword remains.

Full suite now green (`✅ ALL CHECKS PASSED`, exit 0), skipping only stage 3 (the ~1.5h Aeneas production-extraction harness, which is orthogonal to this Lean+docs change and runs as a separate CI job):

- 1/8 cargo test — 73 passed · 2/8 aeneas_extract — passed
- 4/8 `lake build` — 9031 jobs · 5/8 Mem generated-artifact wrapper — passed
- 6/8 V1 — 16/16 · 7/8 V2 — 15/15 (`root_soundness` strong-export closure: 0 `ZiskFv.*` axioms, baseline matches) · 8/8 flake — passed

Both wording fixes (`⟺` → reconstruction-direction-only; dropped vestigial `witnessStore_nondegenerate`) are in `a460d43c`.

PR #231, opened 2026-07-06
Model Main register-prev MemBus pushes

codygunton:
Replacement review PR for the reverted unreviewed merge of #228.

Slice A of the #225 split:
- Adds Main row-local register-prev MemBus push emissions.
- Updates Main row/free/provenance surfaces and downstream row extraction/bridge logic.

Prior local verification before the reverted merge:
- lake build ZiskFv.AirsClean.Main.Circuit
- lake build ZiskFv.AirsClean.FullEnsemble.Balance.RowExtraction
- lake build ZiskFv.AirsClean.FullEnsemble.Balance.MemBusRowBridges
- lake build ZiskFv.AirsClean.FullEnsemble.Balance
- lake build ZiskFv

Do not merge without review approval.

PR #229, opened 2026-07-06
Add register MemBus balance witness

codygunton:
Part 2 of #225, stacked after landed PR #228.

This adds the checked core register MemBus balance artifact for the corrected no-prelude target `add x1,x1,x1`:

- defines the register PIL message shape `[MEMORY_REG_OP, reg, step, 8, value_0, value_1]`
- represents the boundary-shaped contribution explicitly, without adding a `root_soundness` binder or modeling boot/reload as ordinary per-row Main emissions
- lists the x1 timeline messages at steps 0/1/2/3 plus idle x2..x31 boot/reload zero messages
- proves the concrete contribution balances via a generic pull/push-pair combinator
- imports the module from `ZiskFv.lean` so the root build covers it

Important scope note: this is the checked balance core. The trust/consistency wrapper, trust-ledger update, and V1/V2 script integration are intentionally left for the follow-up PR C.

Verification run locally:

- `rtk lake env lean ZiskFv/Compliance/RegisterMemBusBalance.lean`
- `rtk lake build ZiskFv.Compliance.RegisterMemBusBalance`
- axiom check for `ZiskFv.Compliance.RegisterMemBusBalance.addX1X1X1_registerMemBus_balanced`: `[propext, Classical.choice, Quot.sound]`
- full `rtk lake build ZiskFv`

Refs #225.

Issue #225, opened 2026-07-01
Model the register memory-consistency argument (MEMORY_REG_OP / mem_op=3) so the ensemble MemBus balances for real instructions


codygunton:
> [!IMPORTANT]
> Current dependency source of truth: the native GitHub issue dependency graph, not historical prose.
> PR #235 landed a real register-partition checkpoint, but did not close the full original #225 claim.

# Current Status

PR #235 modeled the register-consistency MemBus traffic enough to land the register-partition balance checkpoint with no new project axioms. That was progress, but the original full-close reading of this issue also depended on work now owned elsewhere.

To avoid duplicating scope, the remaining full-close interpretation is represented by native `blockedBy` edges:

- #219 owns whole-channel `witness.BalancedChannels` / accepted-trace construction.
- #169 owns the prev-step timestamp-chain derivation from ordering/range checks.
- #103 and #76 own cross-segment / continuation memory.

# Scope Boundary

#225 no longer blocks #115. #115 has been amended to close at direct-Mem scope with named certificate fields and separate follow-ons for MemAlign / transition-derived Mem source facts.

#225 also should not silently absorb #219, #169, #103, or #76. When those dependencies land, this issue can be closed or reduced to any remaining register-MemBus-specific delta that is not already owned by them.

# Corrections From The Audit Trail

- The register reload push is `main.pil:450` (`reg_pre_load` "Proves the last access."), not the old `main.pil:454` citation.
- The non-degenerate spike target is `add x1,x1,x1`, not `add x0,x0,x0`; ZisK decodes `x0` operands as immediate/no-op accesses emitting no `mem_op=3` traffic.

Historical evidence and implementation notes remain in the issue edit history and comments, especially the PR #235 closeout and the reopening clarification.

codygunton:
Revised the body after a full channel-fidelity audit of the ensemble. Key corrections vs the first draft:

- **The fix is Main-component-local, not a Mem-SM change.** There is no register AIR; `mem_op=3` balances by *telescoping* on the shared MemBus (boot global-pull + Main per-row push-prev/pull-current + per-segment reload). The Mem component (`componentWithDualMemBus`) is not the register provider — don't add register handling there.
- **Spike target = `add x0,x0,x0`**, which telescopes to zero self-containedly (boot seeds value 0, x0 always 0) — no operand prelude needed.
[...]

Issue #219, opened 2026-07-01
root_soundness instantiation: forward BalancedChannels + constraints_hold for the single-ADD witness (#74)


codygunton:
## Goal

Assemble the single-ADD ensemble witness and prove `BalancedChannels` **forward**
— the first forward channel-balance construction in the tree — plus the
full-witness `constraints_hold`. This turns the per-component characterization
lemmas into a real accepted-trace witness.

## Scope

1. Assemble an `EnsembleWitness (fullRv64imEnsemble 1 program).ensemble` using the
   #217 builder: a real ADD Main row + the Binary static-lookup provider row + Mem
   provider rows for the rs1/rs2 reads and rd write + the 8 remaining tables empty.
2. **Forward `BalancedChannels`** (`ens.channels = [MemBus, OpBus]`):
   - OpBus: Main's ADD emit and Binary's ADD provide are the same message with
     opposite multiplicity → cancel; all other tables contribute nil.
   - MemBus: Main's register accesses cancel against the Mem provider entries.
   Use the #218 reduction lemmas + the #217 balance combinator.
3. **`constraints_hold`** for the assembled witness (Main + Binary + Mem rows via
   #218; empty tables vacuous; verifier table via `verifier` = `.empty`).
4. Also discharge `main_height` for the 1-instruction Main table.

## Notes

- This is where the `∀ msg, balanceOf L msg = 0` combinator from #217 gets its
  first real (non-nil) exercise.
- The forward balance produced here is exactly what `stepStrong_<op>` consumes to
  derive the op-bus provider match, so it feeds the row-side proof in the next
  issue.

## Acceptance

- A checked-in `AcceptedZiskTrace 1` (or the witness + the three proof fields)
  for a single ADD, with forward `BalancedChannels` + `constraints_hold` proven.
- No new axioms, no `sorry`; V1+V2 gates green.

Part of #74 (blocks it). Depends on #218 (and #217).

codygunton:
## Gap notes from #225 (register MemBus) — for the #219 non-degenerate witness

#225 landed the register-consistency emissions + a register-partition balance (branch `close-225`). Two things it deliberately leaves to #219, plus one correction:

[...]

PR #232, opened 2026-07-06
Add register MemBus balance witness

codygunton:
Replacement review PR for the reverted unreviewed merge of #229.

Slice B of the #225 split, stacked on #231:
- Adds ZiskFv.Compliance.RegisterMemBusBalance.
- Proves the concrete add x1,x1,x1 register MemBus balance artifact.
- Adds the root import.

Prior local verification before the reverted merge:
- lake env lean ZiskFv/Compliance/RegisterMemBusBalance.lean
- lake build ZiskFv
- axiom check for addX1X1X1_registerMemBus_balanced = [propext, Classical.choice, Quot.sound]

Do not merge without review approval.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -162,6 +162,7 @@ import ZiskFv.Compliance.TraceLevelExport.RomDecodeBindingOps
 import ZiskFv.Compliance.TraceLevelExport.ProgramDecode
 import ZiskFv.Compliance
 import ZiskFv.Compliance.EnsembleWitnessBuilder
+import ZiskFv.Compliance.RegisterMemBusBalance
 -- In-build per-opcode static decode/row-mode pin discharge from the real
 -- Aeneas-extracted ZisK lowerer (eth-act/zisk-fv#111). Standalone; not yet
 -- wired into root_soundness. Per-theorem collectAxioms keeps the aeneas runtime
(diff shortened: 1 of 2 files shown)


codygunton:
**Review — approve, with one scoping point stated plainly.** Reviewed in the #231→#232→#233 stack at tip `e206ce39`.

**Good call on x0→x1.** ZisK decodes `x0` operands as immediate/no-op, so `add x0,x0,x0` (the issue's spike target) emits no `mem_op=3` traffic — a degenerate witness. `x1` is the right minimal real-register target, and the file's note explaining why is worth keeping.

**Message shapes are faithful.** `regMsg reg step v0 v1 = [3, reg, step, 8, v0, v1]` matches the PIL register message; for a register-register ADD the per-row multiplicities are exactly ±1, so modeling with `pulledValue`/`pushedValue` is appropriate.

**The one thing to be clear-eyed about:** `addX1X1X1_registerMemBus_balanced` proves the *generic* fact "a list of matched pull/push pairs balances" — `pairedInteractions_balanced` is universally quantified over `msgs`, so the `addX1X1X1RegisterMessages` content is essentially decorative for provability. The witness builds messages directly via `regMsg`/`pulledValue`/`pushedValue`; it does **not** connect to the real `aRegPreMessageExpr` emissions or real row values from `fullRv64imEnsemble`. So this demonstrates the telescoping *shape*, but #225's actual acceptance criterion — constructing `witness.BalancedChannels` for the real trace — is **not** met here; that remains open (→ #219).

That's not laundering: standalone theorem, kernel-only axiom closure (`propext`, `Classical.choice`, `Quot.sound`), not wired into `root_soundness`, and the PR body honestly calls it "the checked balance core" / "generic pull/push-pair combinator." I just want the stack's readers to not read this as "the ensemble now balances for a real ADD." Useful scaffolding; the real forward-construction is the follow-up.

codygunton:
Addressed review scoping note in 2e552b3: RegisterMemBusBalance.lean now explicitly says the theorem is over listed register MemBus messages and does not yet extract from fullRv64imEnsemble rows / construct trace-level BalancedChannels. Focused check: lake env lean ZiskFv/Compliance/RegisterMemBusBalance.lean.

Issue #226, opened 2026-07-06
Clean Air.Flat can't carry both channels and a transition constraint on one component (fork divergence D1); upstream the fix


codygunton:
## Summary

This is the missing tracking node for a Clean-layer deficiency we hit and worked
around during #100 (PR #163), and for upstreaming the fork patch that fixes it.
The deficiency was **recorded only in a fork-divergence doc on a branch**
(`docs/clean-fork-divergences.md`, entry **D1**, `xcap-100` line) — not as an
issue or PR — which is why it's hard to find. This issue gives it a real node.

## The deficiency

Clean's modern AIR layer, `Air.Flat` (the one zisk-fv uses), is **single-row by
design**. From `Clean/Air/FlatComponent.lean:7-10`:

> "A flat AIR component: one circuit whose constraints are checked independently
> on each row. **There are no direct adjacent-row constraints;** communication
> with other rows/components is expressed by channel interactions."

So an `Air.Flat` `Component` can use **channels** (→ channel balance) but has **no
transition / adjacent-row constraint**. Clean *does* have adjacent-row
constraints — but only in the older, separate, unbridged `Clean/Table` layer
(`InductiveTable` / `CellOffset.next` / `everyRowExceptLast` / `table_soundness`).
The two layers are disjoint (`Clean/Air/README.md`, "Relation To Clean/Table":
*"The two layers are currently independent…"*). Net: **no single component could
carry both channels and a transition constraint.**

## Why it bit us

ZisK's Main AIR enforces a genuine cross-row **polynomial** PC-handshake
constraint (`main.pil:409-410`). It is **not** a channel, so it cannot be modeled
as a per-row `Air.Flat` constraint, nor faithfully re-expressed as a channel —
yet Main also needs channels for the op / mem / rom buses. One component, both
requirements.

## The workaround (already merged, via the fork)

Fork `codygunton/clean`, branch `air-flat-transition-constraints @ 497e4a41`, adds
an **additive, inert** facility to `Air.Flat`:

- `Air.Flat.Component.transition : Input F → Input F → Prop := fun _ _ => True`
  (new field, defaults trivial so existing components/proofs are unaffected);
- `Air.Flat.Table.TransitionConstraints` (holds on each consecutive row pair);
- `Air.Flat.EnsembleWitness.TransitionConstraints` (folds over `allTables`).

On the zisk-fv side the transition is consumed as a verifier-checked accepted-
[...]

PR #236, opened 2026-07-07
#115 direct-Mem read-soundness closeout

codygunton:
## Summary

This is the #115 direct-Mem read-soundness closeout stack.

It derives the accepted-trace-to-boot read-soundness path for direct mutable-Mem accesses, including store rows, on the scoped-direct route. The stack keeps Route A banked, preserves the reviewed scoped-direct commits, adds the consolidated `main_step_index_fixed` certificate class, names the scoped direct replay-length certificate, discharges the active selector residues, marks the new proof predicates reducible where they sit in hypothesis positions, and documents the direct-Mem end state in the trust ledger.

The final integration commit changes the replay guard from `0 < numInstructions` to objective `MutableMemPresent witness`, so memory-less concrete traces construct without weakening the memory-carrying read-soundness path.

## Range

- Base: `main`
- Head branch: `issue-115-readsound`
- Head SHA: `08efda70f4148b19b33a1c4db870661243d9255a`
- GitHub-listed range: 81 commits

## Verification At Head SHA

Run at `08efda70f4148b19b33a1c4db870661243d9255a`:

- `lake build` — passed, 9037 jobs
- `trust/scripts/check-all-semantic.sh` — passed 16/16

- Direct-Mem scope only. MemAlign-family read-soundness remains in #242, gated on MemAlignRom extraction.
- The fixed-column/order certificates are explicit accepted-trace obligations; derivation from range-check fidelity / prev-step chain is #249.
- Mem cross-row derivation through `Component.transition` remains #243, blocked by #226.
- #115 should be closed after merge with the amended-scope closeout comment, not by PR auto-close text.


codygunton (commented):
## Review (Claude, plan author/reviewer session) — checkpoint review against PLAN_READSOUND_115

**Verdict: on track as a checkpoint, with two plan deviations to correct before landing and one concrete technical fix to make now.** Not mergeable in this state — the PR body already says so, and I agree with its own review-focus list. The hygiene is excellent: compiled-checkpoint-only commits, discarded WIP kept off the branch, and a STATUS/plan trail that accounts for every checkpoint honestly.

### Independently verified (I ran these at 463a63f1)

`lake build ZiskFv` green (9036 jobs), V1 16/16, V2 16/16 — a stronger verification state than the PR body claims for this HEAD. One coverage note: V2's global-theorem-binders gate did not flinch because `root_soundness`'s *binder* surface is nominally unchanged — the trust movement here lives in **structure fields** (`AcceptedZiskTrace.mem_replay_source`, `BootSegmentMemorySeed.readSoundInputs`), which that gate does not walk. The audit surface for this refactor is the structure definitions plus `trusted-base.md`, i.e. exactly this review.

### On-plan and done well

- **Stage 1a** done as specified: `AcceptedZiskTrace.memTableRow_specs` derives row-local Mem `Spec` from `spec_holds`, and its module header explicitly refuses to construct the generated replay facts — the right boundary.
- **Placement exhaustiveness** (Stage 2 prerequisite) done: loads/`sd`/narrow stores pinned, all non-memory ops pin `rowsOf i = []`.
- **Stage 0, transfer half** genuinely proven: `memoryBusRowsPrefixReadSound_swap_adjacent` + `memoryBusRowsPrefixReadSound_of_replaySafePermutation` are real proof content, and `readSound_of_bootSegmentReadSoundInputs` reassembles the exact old seed predicate.
- **Stage 2 chain** progressing with residues named in statements, not hidden (`MemProviders.lean`, the split non-mutable branch residues in `MemBusRowBridges.lean`).
- Degenerate witness updated cleanly (guards make all new fields vacuous at n=0).

### Deviation A — sequencing (recoverable; sets the merge bar)

The plan ordered Stage 0's *derivation* half before any public-surface change, precisely so `root_soundness`'s surfaces would not be reshaped around scaffolding that might not close. The seed/trace refactor landed first, and `BootSegmentReadSoundInputs.order` is still an assumption. Acceptable for a review checkpoint; the merge bar is: **derive `order` from accepted-trace facts (timestamps + Mem-table sortedness) for the scoped access class, or bring the re-scope back for explicit sign-off.** Do not land carried-`order` as the "done" state.

### Deviation B — `mem_replay_source` bundling + skipped fork 1b

The plan's Stage 4 required per-class certificate fields with PIL citations, "NOT one bundled callback", and made the cross-row-constraint question an explicit owner sign-off fork (1b). `mem_replay_source` instead carries the entire `FullWitnessMemReplayBridge` as one Σ-bundle on the acceptance definition. The PR flags it for review — here is the ruling:

- Acceptable as **checkpoint scaffolding only**. Before landing: split per factor class, and **do not carry `MemTableGeneratedFixedColumnFacts` at all** — it is deterministically dischargeable via `segmentWithFixedL1` (`TableProjections.lean:1168`), so carrying it is pure trust inflation.
- The 1b decision (derive the cross-row constraint package via a Mem `Component.transition`, vs. carry it as certificate) is still open and is the owner's call — raise it explicitly rather than defaulting to carry-everything.
- State the cost plainly in `trusted-base.md`: this field **strengthens the acceptance definition**; every future `AcceptedZiskTrace` constructor — #219's single-ADD witness first — must now produce the bundle. The current ledger text ("partial trust reduction... residue is structural/boot/order data") is fair about the seed side but silent about this constructor burden. Post the #219 gap note (plan Stage 5) — not yet done.

### Technical finding — the `swap` kernel is too strong to derive `order` later

`MemoryBusRowsReplaySafePermutation.swap` requires byte-disjointness (both directions) **and** the replay-commutation equation. Same-address read/read adjacent pairs — explicitly PIL-legal (equal-timestamp dual read rows; your own infra records this in the `Nodup`-not-required notes) — commute but are **not** byte-disjoint, so the relation as defined cannot reorder them. The Stage-0 derivation will wall exactly there on real traces. Note the fix is *not* to drop conditions down to commutation alone: a write/read pair on the same bytes commutes memory-wise while changing the read's observed prefix, so commutation is insufficient for the transfer. The right side condition is pairwise "**neither row is an active write overlapping the other's read/write bytes**" (byte-disjoint ∨ read/read as special cases). Generalize `memoryBusRowsPrefixReadSound_swap_adjacent` and the constructor now, before building the derivation on top.

### Accuracy nits

- `MemoryBusRowsReplaySafePermutation`'s docstring says it "does not mention read values or replay agreement" — the `swap` constructor literally carries a quantified replay-commutation equation. Say "does not state read-value agreement" instead.
- The relation docstring's framing as "the Stage-0 order-transfer target" is right — keep that sentence; it is the honest statement of what remains.

### Bottom line

Real progress in the plan's direction: the semantic read-value content is now derived, the residues are named and visible, and the proof infrastructure is sound. Finish = derive `order` (after generalizing the swap kernel), split/narrow `mem_replay_source` with the 1b decision made explicitly, ledger + #219 notes, then full gates again.

codygunton (commented):
## Review round 2 (Claude, plan author/reviewer session) — at 3ae03203

**Verdict: every finding from the first review was addressed, and addressed correctly.** Still a checkpoint (the PR says so), but the remaining work is now concentrated where the plan wanted it: one hard derivation, one owner decision, and residue closure.

### Independently verified at 3ae03203

Full `lake build ZiskFv` green, V1 16/16, V2 16/16 — stronger than the PR body's focused-checks claim for this HEAD.

### Round-1 findings — resolution check

1. **Swap kernel generalized (89ad0cc0)** — `MemoryBusEntryByteDisjoint` → `MemoryBusEntryNoActiveWriteOverlap` in both directions. This is exactly the right predicate: an active write must be byte-disjoint from its swap partner; read/read and inactive pairs pass trivially; overlapping write/read and write/write swaps stay excluded. The Stage-0 derivation wall I flagged (PIL-legal same-address read/read dual rows) is gone.
2. **`mem_replay_source` split** — replaced by per-class guarded fields: `mem_replay_table` (pure selection subtype), sidecar source columns (`segment`/`permutation`/`gsum`/`im0`/`im1`), `mem_replay_constraints`, `mem_replay_row_ranges`, segment ranges, and coverage. `MemTableGeneratedFixedColumnFacts` is **derived** via `segmentWithFixedL1` and no longer carried — per the ruling. One ledger nit: state explicitly in `trusted-base.md` that the sidecar *data* columns are pinned by `mem_replay_constraints` (a constructor cannot fabricate them without falsifying the constraint facts) — the field docstrings say this; the ledger should too.
3. **Stage 1b escalated properly** — issue #237 records the cross-row/source surface decision for the owner instead of defaulting. Correct handling.
4. **Beyond the asked fix — two residues *derived* instead of carried:** the Main self-branch exclusion now follows from 14-flag selector booleanity proven from the composed Main constraints (`Main/Circuit.lean`), and the RegisterBoundary branch exclusion follows from `mem_op = 1 ≠ 3`. Both were candidate certificates in the plan; deriving them is the best outcome.
5. **Order residue named** — `BootSegmentReplaySafeOrderCertificate`, with permutation/membership APIs and placement-to-execution-row hooks. Honest and visible.

### One process note — make the Fork A choice explicit

The new MemAlign branch-following work (structural provider-row predicates, `SubdoublewordLoadProviderWitness` conversions with explicit residues) is a *middle course* on the plan's Fork A (Mem-direct-only vs. full MemAlign chain), taken without an explicit sign-off. The engineering choice is defensible — this coverage is needed under any fork outcome and the residues are visible, not buried — but record the chosen scope the same way Stage 1b got #237: a short note on #115 (or a sub-issue) saying which Fork A option this is converging to and what the MemAlign end-state residue will be.

### Remaining before this can land (the bar is unchanged)

1. **Derive `BootSegmentReplaySafeOrderCertificate` from accepted-trace facts** — placement timestamps + Mem-table (addr,step) sortedness. Still the make-or-break item; the generalized kernel and the new placement/order hooks are the right on-ramp, but the derivation has not started.
2. **#237 decision** (owner) — derive the Mem cross-row package via a transition-style surface vs. approve the split certificate residue.
3. **MemAlign residue closure** (or a scoped certificate + the Fork A record above); the load wrapper still carries the subdoubleword replay/value assembly residue.
4. **Store-side correspondence** — only the load `b`-side path is specialized so far; `rowsOf` also pins store `e2` rows.
5. #219 gap note (the new `AcceptedZiskTrace` constructor burden), final ledger regen, full gates at the landing HEAD.

### Bottom line

High-quality response to review: three residues eliminated by derivation rather than carried, the trust surface reshaped into exactly the per-class form the plan required, and the two owner decisions surfaced instead of defaulted (one formally via #237 — do the same for Fork A). The branch remains honest scaffolding until the order certificate is derived.

codygunton:
Review checkpoint requested for #115 direct-Mem closeout at `29221ed`.

Final gates passed: full `lake build` (9037 jobs), V1, V2, `nix run .#test`, explicit axiom scan on the new endpoint/certificate surface, and the read-soundness-premise grep. The PR body has been refreshed with the owner decisions, landed commits, verification, and follow-on residues (#219/#242/#243/#225).

codygunton (commented):
## Final review (Claude, reviewer-of-record) — guard commit 08efda70 + landing state

**Approved for merge.** The `MutableMemPresent` guard surface implements the authorized shape exactly: an objective Prop over the witness (a mutable-Mem table with rows exists), all `mem_replay_*` fields re-guarded on it instead of `0 < numInstructions`, and the seed carrying the honest bridge `memPresent_of_executionRows_nonempty` (a trace whose execution rows are nonempty must present the Mem bridge) — so memory-less witnesses discharge everything vacuously while memory-carrying traces retain byte-for-byte the prior obligations. No endpoint conclusions changed. The forcing incident (first memory-less concrete witnesses) is recorded in the body as directed.

Independent verification at 08efda70 (via the stacked f502c9f1, which contains it): full `lake build`, V1 16/16, V2 16/16 — re-run by the reviewer in addition to the agent's per-head full-gate runs including `nix run .#test`.

This PR is the #115 closeout: the amended acceptance criteria are met (read-soundness derived, store-inclusive, direct-Mem scope; residues named and filed as #242/#243/#249). Merging in the handoff order.

Issue #220, opened 2026-07-01
root_soundness instantiation: apply root_soundness to a concrete ADD + multi-row ALU trace (headline of #74)


codygunton:
## Goal

The headline OPEN criterion of #74: a checked-in instantiation of
`ZiskFv.Compliance.root_soundness` (`ZiskFv/Soundness.lean`) on a concrete trace
fragment, through the current surface (`AcceptedZiskTrace` / `SailTrace` /
`ZiskStep` / `RowDecode` / `InputsAgree` / `RowOutsideDefectRegion`) — not the
internal `zisk_riscv_compliant_program_bus`.

## Scope

1. Build the row-side evidence over the ADD witness from #219:
   - `sailTrace : SailTrace 1` (a plain `Fin 1 → SequentialState`);
   - `ziskStep : ZiskStep.add` carrying `Claim_add`;
   - `rowDecodes : Decode_add` — discharge the `mainOfTable trace.program
     trace.mainTable` field equalities (op / is_external_op / m32 / store_pc /
     exec-row shape);
   - `inputsAgree : Inputs_add` — the register reads, PC, lane bridges,
     `h_nextPC_matches`, rd index;
   - `hAvoidKnownBugs : RowOutsideDefectRegion` = `True` for the add arm.
2. Apply `root_soundness` to obtain `∀ i : Fin 1, StepSound …`.
3. **Extend to a multi-row non-memory trace** (e.g. `ADDI; ADD; SUB`) so the
   `∀ i` conclusion dispatches across a heterogeneous `ZiskStep` sequence and the
   witness/balance machinery is exercised past one row. Non-degenerate on the
   compute side.

## Current route

#219 provides the base single-ADD accepted trace. PR #248 implements the current-surface #220 route as an `ADD + JAL x0,0` spin-loop trace: the ADD row falls through to pc 4, the final executed JAL self-loops at pc 4, and the successor Main row reuses the committed JAL ROM row. This avoids the exact-ROM successor-row blocker without changing `root_soundness`, `Decode_*`, or `AcceptedZiskTrace`.

#245 remains a realistic-programs follow-up, but no longer blocks #220's spin-loop route.

## Acceptance

- A checked-in file (under the `root_soundness_instantiation_*.lean` glob from
  #217) applying `root_soundness` to a concrete single-ADD trace, and to a
  2–3-instruction ALU trace.
- Header documents the program + how to regenerate the literals.
[...]

PR #235, opened 2026-07-07
#225: model register MemBus (MEMORY_REG_OP) so the ensemble balances for real instructions

codygunton:
Closes #225 (register-partition close).

## What this does

The full RV64IM Clean ensemble previously **could not balance its MemBus for any real instruction**: every active Main row emits `MEMORY_REG_OP` (`mem_op=3`) register pulls, and the composed providers emit only `mem_op∈{1,2}`, so register traffic was unbalanceable. This PR models the register memory-consistency argument — a telescoping chain on the shared MemBus — as **derived composed-table emissions** (no new premise, `axiom`, or `opaque`), and proves the `mem_op=3` partition balances for `add x1,x1,x1`.

### Stage A — RegisterBoundary provider component (`ZiskFv/AirsClean/RegisterBoundary.lean`)
New single-purpose Clean provider emitting, per tracked register, the two chain-closing boundary terms ZisK emits outside Main's per-row access loop:
- **boot pull** — `global_init_mem` (`mem.pil:507-508`, per-register call site `main.pil:535-537`): `mem_op=3`, ts 0, value 0, mult −1.
- **reload push** — `reg_pre_load` "Proves the last access." (`main.pil:450`): ts = last access, mult +1.

`Assumptions := True`, no algebraic constraints — a deliberately under-constrained boundary provider, which is sound in the soundness direction (an under-constrained provider never vacuates `root_soundness`; the anti-laundering hazard runs the other way).

### Stage B — compose into `fullRv64imSoundEnsemble` + thread the cascade
Composed via one `addTable`. The balance-classification chain gains the new provider:
- `Classification`: `component_mem_fullRv64im_cases` + a new `registerBoundary_table_interactionsWith_opBus_nil`.
- `CounterpartClassification`: OpBus lemma dismisses it (no OpBus); MemBus lemma adds it as a genuine provider.
- `MemBusRowBridges`: the self-contained #219 provider-classification chain carries a bare **structural** RegisterBoundary provider branch (new `ActiveMainRegisterBoundaryProviderRowMatchSpec`). These are **partition-disjointness passthroughs** (`mem_op=3 ∉ {1,2}`), **not** new register-timeline reasoning in the data-memory proofs — a concrete data-memory main discharges them by disjointness, which the #219 instantiation supplies.

Degenerate n=0 witness re-verified unchanged (new table 0 rows).

### Stage C — register-partition balance from real emissions (`RegisterMemBusBalance.lean`)
[...]

PR #244, opened 2026-07-09
Construct single-ADD accepted trace

codygunton:
## Summary

This PR delivers the #219 single-ADD accepted-trace construction stacked on #236.

It extracts the register MemBus partition from actual Main/RegisterBoundary `interactionsWith` reductions, assembles the concrete single-ADD ensemble witness, proves `constraints_hold`, `BalancedChannels`, transition/main-height/fixed-column side obligations, and packages `singleAddAcceptedTrace`. The final stacked repair fills the #236 `MutableMemPresent` guard fields by proving the single-ADD witness has no mutable-Mem rows instead of fabricating memory replay evidence.

## Range

- Base: `issue-115-readsound` (`#236`)
- Head branch: `issue-219-single-add`
- Head SHA: `fc651e05009d0b5e652cd2d53472fae871180eac`
- GitHub-listed range: 10 commits

## Verification At Head SHA

Run at `fc651e05009d0b5e652cd2d53472fae871180eac`:

- `lake build` — passed, 9038 jobs
- `trust/scripts/check-all-semantic.sh` — passed 16/16

- Stacked on #236; merge after #236.
- Single-ADD accepted trace only. It does not instantiate `root_soundness`; that is #248.
- #225 review should confirm whether the reopened criteria are fully met; the remaining ordering/range-chain criterion is tracked by #249.
- Multi-row ALU trace construction remains in #220 after #248.


codygunton (commented):
## Review (Claude, reviewer-of-record) — full audit of the 9-commit range

**Verdict: approve after one cheap must-fix.** This is the crux artifact of the #74 ladder — the first non-degenerate `AcceptedZiskTrace` — and the audit found it honest throughout. Gates independently verified green (build + V1 16/16 + V2 16/16) at the stacked #248 head, which contains this range.

### What the audit confirmed

- **Genuine forward balance on both channels**: MemBus reduces the *actual* `flatMap` of all 11 tables' `interactionsWith` to the RegisterBoundary (31 rows) + Main (6 emissions) list and telescopes it; OpBus cancels Main's push against the BinaryAdd provider's pull on the same message. Multiplicities are the emissions' own selectors, table-extracted — this also retires the "not yet extracted via `interactionsWith`" disclaimer from the #235 review, the preferred fix landing at last (note this on #225).
- **Genuine constraints**: per-table `ConstraintsHold` from explicit `ProverAssumptions`, not blanket `decide`; empty tables via the empty-table lemma.
- **Honest vacuities, disclosed**: `transitions_hold` vacuous at length 1; the within-segment half of `segment_l1_fixed` vacuous; `main_height` and first-row `segment_l1` genuinely proven.
- **No check-tuning**: register pointers raw (x1 = 1 everywhere), prev-step chain 0/1/2 with currents 1/2/3, boot 0 / reload 3, `op = OP_ADD`, `jmp_offset = 4/4` (the 056c66f4 fix moved *toward* the real lowering). No new axiom/sorry/opaque/native_decide in the range.
- Non-witness surface touched: one visibility widening (`toElements_eval_toArray` private→public) and the `mainRegisterInteractions` re-pointing to table-extracted form with an `_eq_` bridge — both fine.

### Must-fix before merge

**No `#print axioms` artifact and no gate wiring exist for this witness.** `singleAddAcceptedTrace` / `singleAddWitness_balancedChannels` are referenced only by their own file — axiom cleanliness is asserted, not perpetually certified. Add a `trust/consistency/` wrapper (naming it `root_soundness_instantiation_*.lean`-adjacent or extending the witness glob) with `#print axioms` on both, so V2 re-checks it forever, as it does the degenerate instantiation. One small file.

### Honest-scope caveats to keep visible (already in the body's scope notes — keep them at merge)

The timestamp/ordering values are pinned, not enforced: every range/ordering table is empty in this witness, so the prev-step chain is asserted-consistent (the #169 axis enforces it in real ZisK). Single-row scope makes the cross-row certificates vacuous. Both are correctly disclosed; neither is a defect of this PR.

Issue #245, opened 2026-07-09
Split accepted trace execution length from committed ROM length for padded Main rows


codygunton:
## Problem

#220 exposed a real surface mismatch for realistic non-looping programs: the current Lean surface ties executed trace length and committed ROM/program length together.

- `root_soundness` takes `AcceptedZiskTrace numInstructions` and `SailTrace numInstructions`.
- `AcceptedZiskTrace numInstructions` carries `program : ZiskInstructionRom.Program numInstructions`.
- `Program length := Fin length -> ZiskRomMessage FGL`, so an `AcceptedZiskTrace 1` has exactly one committed ROM row.

That exact-program surface is too narrow for a realistic program whose final executed instruction falls through to a padded successor Main row at a PC that is not itself a committed ROM row.

## Fidelity finding from #220

The successor row cannot be an arbitrary selector-zero row:

- `main.pil:409-410` makes the non-first Main-row PC follow the previous row's next-PC mux.
- For a 4-byte ADD, the faithful next PC is `4`; real register-op lowering uses `zib.j(inst_size, inst_size)`, so ADD's `jmp_offset1/jmp_offset2` are `4/4`.
- Main's ROM lookup is unconditional in the modeled per-row constraints. A successor row with `pc = 4` therefore needs a committed ROM entry at that PC.
- `main.pil:461` notes all-zero padding ROM rows are prevented from matching, so treating this as an unconstrained padding lookup would be dishonest.

A separate #219 repair (`eth-act/zisk-fv#244`, commit `056c66f`) fixed the concrete ADD row/program offsets from `0/0` to faithful `4/4`.

## Current status

This no longer blocks #220. PR #248 takes the owner-approved spin-loop route: execute `ADD x1,x1,x1` at pc 0, then execute `JAL x0,0` at pc 4, with the successor Main row reusing the same committed JAL ROM row. That closes #220's current-surface ADD instantiation without changing `root_soundness`, `Decode_*`, or the accepted-trace/program-length surface.

This issue remains the follow-up for realistic non-looping programs, where the faithful successor/padding row cannot reuse a self-loop ROM entry and the accepted execution length vs committed ROM length mismatch must be represented directly.

## Acceptance

Owner decision pending on the exact surface shape for realistic non-looping programs. Candidate direction: split accepted execution length from committed ROM/program length, or otherwise represent padded successor Main rows without pretending all padding rows have committed ROM entries.

## Scope boundary

[...]

PR #234, opened 2026-07-06
Add concrete row reductions for channel balance (#218)

codygunton:
## Development Milestone

MVP / soundness track for the root_soundness instantiation ladder (#74): provide row-local concrete interaction and constraint reductions needed by the forward BalancedChannels witness.

Closes #218.
Unblocks #219.
Depends on #217, already closed.

## Summary

- Add ZiskFv.Compliance.Instantiation.ConcreteRowReductions.
- Reduce a concrete Main one-row table OpBus emission to the explicit numeric-array interaction used by the ADD witness.
- Reduce concrete Binary staticLookup OpBus provider and Mem dual MemBus provider rows to explicit interaction arrays.
- Add per-row Table.Constraints facts for Main, Binary staticLookup, and Mem rows, parameterized by honest component ProverAssumptions.
- Import the reductions through ZiskFv.lean for downstream use.

## Scope Notes

- The PR is intentionally row-local characterization scaffolding for #219, not the final trace-level BalancedChannels witness.
- Main register-MemBus balancing was reshaped and landed through #231/#232/#233; this PR keeps the Main-side reduction OpBus-focused and leaves final Main register-MemBus integration to #219.
- The original duplicate degenerate root_soundness witness fix was dropped during review resolution because current main already carries that fix.

## Review Resolution

The agent review blockers have been addressed:

- Rebased onto current origin/main at 3ac200ac.
- Dropped the redundant degenerate witness commit; this PR no longer touches trust/consistency/root_soundness_instantiation_degenerate.lean.
- Added the five register-MemBus MainRomRow fields to mainSpikeRow with zero values.
- Added the module-doc pointer to the #219 consumer.
- Current merge state is CLEAN at 5bccb814.

## Verification

- lake env lean ZiskFv/Compliance/Instantiation/ConcreteRowReductions.lean
- lake build ZiskFv.Compliance.Instantiation.ConcreteRowReductions
- lake build
- trust/scripts/check-all.sh
- trust/scripts/check-all-semantic.sh
- git diff --check
- forbidden-token scan on ConcreteRowReductions.lean and ZiskFv.lean
- public lemma #print axioms scan: only propext, Classical.choice, Quot.sound

PR #160, opened 2026-06-27
#111: discharge aeneasBridgeTrust — all 63 RV64IM static decode pins proven in-build from the real Aeneas lowering (kernel-sound)

codygunton:
## What this does

Proves, **in the main `lake build`** and kernel-soundly, that ZisK's *real* Rust→ISA lowerer
(faithfully extracted by Aeneas into `trust/aeneas/ProductionM2.lean`) produces the **static decode
pins** (`op`, `is_external_op`, `m32`, `set_pc`, `store_pc`) that the per-opcode soundness proofs name
— for all **63 RV64IM opcodes**, at the **top-level dispatcher** level. This replaces the previously
*out-of-band* (`AENEAS_CHECK_RV_COMPLETENESS`, `native_decide`) decode-fidelity check with in-build
kernel theorems.

Module: `ZiskFv/Compliance/AeneasBridgeTrust/Extraction/` (`Helpers` + family files), imported by
`ZiskFv.lean`. **Two layers:**
- **entry-point lemmas** (`<op>_static_pins` / `<op>_extracted_rowMode_pins`) — facts about the
  builder entry points (`create_register_op_typed`, `immediate_op_typed`, `load/store_op_typed`,
  `lui`/`auipc`/`jal`/`jalr`/`nop`/`copyb`); the reusable proof core.
- **dispatcher theorems** (`Extraction/Dispatch.lean`: `<op>_dispatch_static_pins` /
  `<op>_dispatch_extracted_rowMode_pins`, 126 theorems) — go through the **real dispatcher**
  `riscv2zisk_single_row.Riscv2ZiskContext.lower_rv64im_single_row_input`, state the **exact routing
  side condition**, discharge the dispatcher's degenerate branches, and reduce to the entry-point
  lemma. Side conditions are explicit, e.g.: register ALU/M ops and branches and most shift-imm ops —
  unconditional; `ADD` — `input_precompile=none ∧ rd,rs1,rs2 ≠ 0`; `OR` — `rs1,rs2 ≠ 0`; `ADDI` —
  `rd,rs1,imm ≠ 0`; `XORI/ORI` — `rs1 ≠ 0`; `AUIPC/JAL/JALR` `store_pc=true` — nonzero rd. (These are
  the honest "degenerate input ⇒ ZisK lowers to a no-op/copyb/DMA row" conditions.)

## Soundness

- Every theorem's `#print axioms` = exactly `[propext, Classical.choice, Quot.sound]` — **no
[...]

PR #248, opened 2026-07-09
#220: instantiate root_soundness for ADD spin-loop trace

codygunton:
## Summary

This PR advances #220 by instantiating `root_soundness` on a concrete memory-less `ADD + JAL x0,0` spin-loop trace, stacked on #244.

It builds the ADD spin-loop accepted trace, proves the no-memory boot/read-soundness inputs under #236's `MutableMemPresent` surface, repairs the trace-level register pointer contract to raw register numbers with Rust/Aeneas/PIL evidence, and adds the add-spin root-soundness consistency witness discovered by V2.

## Range

- Base: `issue-219-single-add` (`#244`)
- Head branch: `issue-220-root-soundness-add`
- Head SHA: `f502c9f1015b716978066fb88f1e6883dccb674a`
- GitHub-listed range: 6 commits

## Verification At Head SHA

Run at `f502c9f1015b716978066fb88f1e6883dccb674a`:

- `lake build` — passed, 9040 jobs
- `trust/scripts/check-all-semantic.sh` — passed 16/16; add-spin root-soundness witness has the expected inherited Sail/root-soundness axiom surface, while `singleAddAcceptedTrace` and `singleAddWitness_balancedChannels` remain at `propext` / `Classical.choice` / `Quot.sound`

- Stacked on #244; merge after #244.
- #220 remains open after this PR for the multi-row ALU slice.
- The realistic non-looping program/successor-row surface finding remains #245; this PR intentionally uses the zero-surface-change spin-loop route.
- The multi-row ALU extension is not mechanical from this branch: it needs multi-row BinaryAdd/provider assembly plus register MemBus telescoping balance.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -168,6 +168,7 @@ import ZiskFv.Compliance.RegisterMemBusBalance
 import ZiskFv.Compliance.Instantiation.ConcreteRowReductions
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
+import ZiskFv.Compliance.AddSpinRootSoundness
 -- In-build per-opcode static decode/row-mode pin discharge from the real
 -- Aeneas-extracted ZisK lowerer (eth-act/zisk-fv#111). Standalone; not yet
 -- wired into root_soundness. Per-theorem collectAxioms keeps the aeneas runtime
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -167,6 +167,7 @@ import ZiskFv.Compliance.EnsembleWitnessBuilder
 import ZiskFv.Compliance.RegisterMemBusBalance
 import ZiskFv.Compliance.Instantiation.ConcreteRowReductions
 import ZiskFv.Compliance.SingleAddWitness
+import ZiskFv.Compliance.AddSpinWitness
 -- In-build per-opcode static decode/row-mode pin discharge from the real
 -- Aeneas-extracted ZisK lowerer (eth-act/zisk-fv#111). Standalone; not yet
 -- wired into root_soundness. Per-theorem collectAxioms keeps the aeneas runtime
--- a/ZiskFv/Compliance/AddSpinRootSoundness.lean
+++ b/ZiskFv/Compliance/AddSpinRootSoundness.lean
@@ -1,6 +1,8 @@
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Soundness
 
+set_option maxRecDepth 10000
+
 /-!
 # Concrete `root_soundness` instantiation for the ADD + spin-loop trace (#220)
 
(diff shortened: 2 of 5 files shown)


codygunton (commented):
## Review (Claude, reviewer-of-record) — 4 commits, stacked on #244

**Verdict: approve after the same one must-fix as #244 (gate wrapper).** This PR is the milestone: `addSpinRootSoundness` is the **first non-vacuous end-to-end application of `root_soundness`** — the headline theorem fired on a concrete 2-instruction trace (ADD x1,x1,x1 then a self-loop jump), with a real `SailTrace`, concrete decode claims, a boot seed whose read-soundness is honestly vacuous for a no-memory trace, and a named `StepSound` corollary for the ADD. Gates independently verified at this head: full build, V1 16/16, V2 16/16.

### The register-pointer contract fix (0f3e33c2) — exemplary handling of a protected surface

The ruling's investigation was done and is cited in the docstring: the Rust lowerer writes raw registers into `store_offset` (`riscv2zisk_context.rs:688-693`, `zisk_inst_builder.rs:289-303`), the PIL uses that column directly as the register bus pointer (`main.pil:316-319`), and boot seeds raw addresses `1..31` (`main.pil:535-537`, `zisk_registers.rs:122`). So `Transpiler.ind = 4·rd` was a hand-written shim bug — notably one the #160 extraction-backed pins never shared. Every consumer compiled *unchanged* against the flipped convention, which is exactly why the bug stayed latent: all internal proofs were convention-agnostic (roundtrip lemmas hold either way), and only a concrete witness forced a discriminating value. The fix is isolated to its own commit, the two closure baselines are regenerated in-commit, and this review is the CODEOWNER review of that diff. This is the anti-vacuity ladder doing precisely what it exists to do — #220 has now flushed out two latent surface defects (successor-row conflation, pointer convention) that no amount of green builds could have found.

Also noted: the spin-loop route resolves the successor-row problem without any surface change (the JAL's successor row reuses its own ROM entry), which parks #245's executed-vs-ROM-length split as a *realistic-programs* follow-up rather than a #220 blocker — update #245 accordingly.

### Must-fix before merge

Add `trust/consistency/root_soundness_instantiation_add_spin.lean` with `#print axioms addSpinRootSoundness` (expected kernel-only) — the existing V2 glob `root_soundness_instantiation_*.lean` will pick it up automatically. Without it, the milestone's axiom cleanliness is a one-time claim instead of a standing gate. Cover #244's witness in the same file or a sibling.

### Review depth note

The 823-line `AddSpinWitness.lean` was reviewed at the machinery level (it reuses the #234/#244 reduction patterns audited in the #244 review) plus gates, not row-by-row; the pointer-fix, instantiation file, and trust surfaces were reviewed in full. Given the reuse and the green semantic gates, that is sufficient for a checkpoint approval; a row-level pass rides with the #221 review when memory rows enter.

### For the record

Remaining on #220's acceptance after this merges: the multi-row ALU extension (more than one "real" instruction before the spin) — mechanically the same construction. #221 (load with real memory prefix) is where the next hard content lives.

Issue #249, opened 2026-07-09
Derive the memory timestamp/ordering certificates from range-check fidelity (step-index, prev-step chain, Mem ranges)


codygunton:
## Summary

Three certificate families on the trust surface share one discharge path — the range/fixed-column
facts behind `STEP = main_segment·N + SEGMENT_STEP` (main.pil) — and none of them currently has a
tracking home. Their discharge paths exist only in docstrings. This issue is that home.

## The carried certificates to derive

1. **`AcceptedZiskTrace.main_step_index_fixed`** — Main's `main_step` column equals the row/trace
   index. Its docstring targets the `SEGMENT_STEP` fixed-column + range facts; precedent for the
   derivation shape is `segment_l1_fixed` (fixed-column class).
2. **The within-row previous-step chain** — `a/b/c_reg_prev_mem_step = 0/1/2`-style values are
   hand-pinned in every concrete witness (single-ADD, ADD-spin). Real ZisK enforces the chain via
   the ordering/range checks (`main.pil:447`-region range_check on `last_segment_reg_mem_step -
   last_reg_mem_step - 1`, and the per-access prev-step range checks). Deriving it removes the
   pins from witness constructors and closes the last criterion on #225.
3. **`AcceptedZiskTrace.mem_replay_row_ranges` / `mem_replay_segment_ranges`** — the Mem-table
   increment/addr/step and distance range facts (`mem.pil:384-385, 397, 267-268`), currently
   caller-supplied because no range static-lookup provider is composed into
   `fullRv64imSoundEnsemble` (the static tables exist in `ZiskFv/AirsClean/RangeTables.lean`;
   the composition does not).

## Why one issue

All three reduce to the same modeling work: compose the range-lookup facts (static-table route,
as Binary/BinaryExtension already do) and the step fixed-column facts into the Clean ensemble, then
replace each certificate field with a derivation. This is the in-repo half of the
`AcceptedZiskTrace` certificate burn-down (the other half is #243, gated on upstream #226).

## Acceptance

- Each of the three certificate families is either derived (field deleted from
  `AcceptedZiskTrace` / witness constructors) or explicitly re-scoped with owner sign-off.
- Witness constructors (degenerate, single-ADD, ADD-spin, and successors) no longer hand-pin the
  prev-step chain.
- No weakening of any existing obligation; 0 new project axioms; V1/V2/`nix run .#test` green;
  `trust/trusted-base.md` updated in the same PR.

## Relations

[...]

PR #250, opened 2026-07-10
#220: instantiate heterogeneous multi-row root soundness

### LINKED ISSUE #220: root_soundness instantiation: apply root_soundness to a concrete ADD + multi-row ALU trace (headline of #74)

codygunton:
## Summary
- generalize Main MemBus reductions and preserve real emission metadata for multi-row register telescopes
- construct an accepted `ADD x1,x1,x1; ADDI x1,x1,0; JAL x0,0` trace with four Main rows and two BinaryAdd rows
- bind all three executed rows to Sail, instantiate `root_soundness`, and add a trust consistency witness

Closes #220

## Verification
- `lake build ZiskFv.Compliance.AddAddiSpinWitness`
- `lake build ZiskFv.Compliance.AddAddiSpinRootSoundness`
- `trust/scripts/check-all.sh`
- `trust/scripts/check-all-semantic.sh`
- `lake build`
- `nix run .#test`

### DIFF
--- a/ZiskFv/Compliance/AddAddiSpinRootSoundness.lean
+++ b/ZiskFv/Compliance/AddAddiSpinRootSoundness.lean
@@ -1,0 +1,509 @@
+import ZiskFv.Compliance.AddAddiSpinWitness
+import ZiskFv.Soundness
+
+set_option maxRecDepth 10000
+
+/-!
+# Concrete `root_soundness` instantiation for the ADD/ADDI/spin trace (#220)
+
+This file binds the heterogeneous accepted Zisk trace to a three-state Sail trace and applies the
+public `root_soundness` theorem at all three executed rows.
+-/
+
+open Goldilocks
+open ZiskFv.Compliance
+open ZiskFv.Compliance.AddAddiSpinWitness
+open ZiskFv.Compliance.AddSpinWitness
+open ZiskFv.Compliance.Instantiation
+open ZiskFv.Compliance.RegisterMemBusBalance
+open ZiskFv.Compliance.RomDecodeBinding
+open ZiskFv.Compliance.SingleAddWitness
+open ZiskFv.ZiskCircuit.MemTrace
+open ZiskFv.Trusted
+
+namespace ZiskFv.Compliance.AddAddiSpinRootSoundness
+
+def x0 : regidx := regidx.Regidx (0#5)
+
+def x1 : regidx := regidx.Regidx (1#5)
+
+def addAddiSpinAddIndex : Fin 3 := ⟨0, by decide⟩
+
+def addAddiSpinAddiIndex : Fin 3 := ⟨1, by decide⟩
+
+def addAddiSpinJalIndex : Fin 3 := ⟨2, by decide⟩
+
+def addAddiSpinAddClaim : Claim_add addAddiSpinAcceptedTrace addAddiSpinAddIndex where
+  r1 := x1
+  r2 := x1
+  rd := x1
+
+def addAddiSpinAddiClaim : Claim_addi addAddiSpinAcceptedTrace addAddiSpinAddiIndex where
+  r1 := x1
+  rd := x1
+  imm := 0#12
+
+def addAddiSpinJalClaim : Claim_jal addAddiSpinAcceptedTrace addAddiSpinJalIndex where
+  imm := 0#21
+  rd := x0
+
+def addAddiSpinZiskStep : ∀ i : Fin 3, ZiskStep addAddiSpinAcceptedTrace i
+  | ⟨0, _⟩ => .add addAddiSpinAddClaim
+  | ⟨1, _⟩ => .addi addAddiSpinAddiClaim
+  | ⟨2, _⟩ => .jal addAddiSpinJalClaim
+
+def addAddiSpinMisa : RegisterType Register.misa := 0#64
+
+def addAddiSpinRegs (pc : BitVec 64) : Std.ExtDHashMap Register RegisterType :=
+  let regs0 := (default : PreSail.SequentialState RegisterType Sail.trivialChoiceSource).regs
+  let regs1 := regs0.insert Register.PC pc
+  let regs2 := regs1.insert Register.misa addAddiSpinMisa
+  regs2.insert (reg_of_fin (regidx_to_fin x1)) (0#64)
+
+def addAddiSpinState (pc : BitVec 64) :
+    PreSail.SequentialState RegisterType Sail.trivialChoiceSource :=
+  { (default : PreSail.SequentialState RegisterType Sail.trivialChoiceSource) with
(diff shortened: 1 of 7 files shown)


codygunton (commented):
## Review (Claude, reviewer-of-record) — 3 commits, Stream 1 of the fanout closeout

**Verdict: approve, no must-fix.** This closes #220's remaining criterion: a heterogeneous multi-instruction `root_soundness` instantiation (`ADD x1,x1,x1; ADDI x1,x1,0; JAL x0,0`, four Main rows, two BinaryAdd rows) with all three executed rows bound to a concrete Sail trace and per-op `StepSound` corollaries. Gates independently verified at `ab6e8594`: warm `lake build` (9040 jobs), V1 16/16, V2 16/16 (the new `trust/consistency/root_soundness_instantiation_add_addi_spin.lean` is picked up by the glob), and axiom closures checked directly — `addAddiSpinRootSoundness` carries exactly the landed add-spin surface (Sail platform/softfloat externs + kernel axioms), while the accepted trace and channel balance are kernel-only. Zero project axioms; no escape hatches in the new files.

### What the audit confirmed

- **The general N-access register telescope is the real deliverable.** `registerAccessChain` / `registerTelescopingInteractions` model the *actual* boundary-first emission order (boundary pulls first state and pushes last; each Main access pushes-previous / pulls-current), and `registerTelescopingInteractions_perm` proves it permutes to paired pull/push for arbitrary N — balance follows generically. The witness applies it at N=5 (ADD's a/b/c + ADDI's a/c on x1), with ADDI's imm-sourced b lane handled by the zero-selector lemma rather than a fake pairing. This is the combinator the plan's spike demanded, and it replaces hand-pinned pairwise reasoning going forward.
- **Balance starts from real emissions.** `addAddiSpinWitness_memBus_balanced` reduces the actual `tables.flatMap (·.interactionsWith MemBusChannel.toRaw)` through proven equalities to the telescope + idle pairs + inactive JAL rows; nothing is hand-authored at the interaction level. The `pairedInteraction` change from Clean's `pulledValue` to the new `emittedPulledValue` is a fidelity *improvement* — the paired lists now literally equal real component emissions including `assumeGuarantees := false`, which is what lets the `_eq` lemmas close definitionally. Reviewed for hidden-promise risk: it is interaction data, not a Prop; fine.
- **Subsumption rule honored at the API level.** `binaryAddSingleRowTable` and its three lemmas are deleted tree-wide (all consumers migrated to `binaryAddRowsTable [row]`), and the private duplicate `zeroInteractions_balanced` is unified into `RegisterMemBusBalance`. `mainMemBusInteractionsFor` looks like a pass-through but is a deliberate unfold barrier over the six-fold `.eval env` definition (prevents kernel normalization in the list-recursion lemmas) — reviewed and justified.
- **Honest scope, correctly held.** No memory in the trace, so `MutableMemPresent` is refuted and every `mem_replay_*` field discharges by refutation — the designed guard shape, disclosed in the file. The prev-step chain values (`a_reg_prev_mem_step := 3`, `store_reg_prev_mem_step := 5`, reload 7) remain hand-pinned — correctly NOT smuggled into a derivation here; that is #249 / Stream 3. `MainStepIndexFixedFacts` is instantiated with `main_step` = row index. Protected surfaces (`root_soundness`, `Decode_*`, `AcceptedZiskTrace`, trust policy files) untouched.

### Notes for follow-up (not blocking)

1. **Residual special-case routes.** SingleAddWitness still balances via the `decide`-based concrete `addX1X1X1_registerMemBus_balanced` chain, and AddSpinWitness keeps an inline zero-selector proof for its JAL rows instead of the new `MainMemBusInactive` route. Both are still consumed, so nothing is dead — but the general lemmas subsume them. Since Stream 2 (#245) must rebase all witnesses anyway, migrate these to the general telescope / inactive-row routes there and delete the concrete chain.
2. **Data-degenerate values.** All register values are zero (0+0=0; ADDI imm=0), so heterogeneity is real at decode/dispatch level but the arithmetic content repeats #244's. Not a #220 criterion — but a nonzero immediate (e.g. `ADDI x1,x1,5`) in the Stream 2 witness rebase would make the anti-vacuity claim strictly stronger at near-zero cost.

### Review depth note

The telescope combinator, `emittedPulledValue` semantics, provider generalization, subsumption sweep, trust file, and all new public definitions were reviewed in full; the 1069-line witness file was reviewed at the machinery level (it reuses the audited #244/#248 reduction patterns) plus the row-value consistency enforced mechanically by the `_eq` lemmas and gates.

Issue #103, opened 2026-06-16
Closed: cross-segment Mem seam capability banked for #76 follow-ons


codygunton:
## Current status (2026-06-24)

Closed as a banked capability spike, not as a live branch to merge.

The useful result was the channel-level cross-segment Mem seam idea and its non-vacuity checks. The old spike branch and intermediate comments predate the current trace terminology and should not be used as current implementation guidance.

Current memory follow-ons:

- #76 is closed at the load-side `LoadMemoryTimelineCoherenceEvidence` / `RowTraceCoherence` floor.
- #115 tracks eliminating or justifying that floor.
- #119 tracks the analogous store RMW byte-residual reduction.

codygunton:
Superseded spike note (2026-06-24): this early extraction-route sketch is historical. #103 is closed as a banked cross-segment Mem seam capability, with current follow-ons in #76, #115, and #119.

codygunton:
Superseded migration note (2026-06-24): this scoped-increment comment predates the current trace/file terminology. Treat it as archival; #103 is closed as a banked capability and is not live implementation guidance.

codygunton:
Superseded make-or-break note (2026-06-24): this records an intermediate #103 spike result on a branch/worktree layout that is no longer the current tree. The closed #103 body is the current summary; follow-ons are #115 and #119.

codygunton:
Superseded caveat note (2026-06-24): this was part of the closed #103 seam spike and referenced old branch-local paths. It is archival, not current implementation guidance.

codygunton:
Superseded Step 2a note (2026-06-24): this intermediate migration sketch predates the current trace/file terminology. Treat it as archival; #103 is closed as a banked cross-segment Mem seam capability, not live implementation guidance.

codygunton:
Superseded framework-wall note (2026-06-24): this was an intermediate #103 spike state with old framework/path references. #103 is closed as banked capability; use the issue body for current status.

codygunton:
Superseded resolution note (2026-06-24): this intermediate #103 comment referenced old branch-local files. The current closed status is recorded in the #103 body.

codygunton:
Superseded landing note (2026-06-24): this was part of the closed #103 seam spike and used old shorthand file references. It is archival; current memory follow-ons are #115 and #119.

codygunton:
[...]

PR #252, opened 2026-07-10
#245: split executed steps from committed ROM length

### LINKED ISSUE #245: Split accepted trace execution length from committed ROM length for padded Main rows

codygunton:
Closes #245.

## Summary
- Separate executed-step count from committed ROM length in `AcceptedZiskTrace`.
- Keep Main/ROM lookups and decode coverage committed-ROM indexed.
- Add an `exec = 1`, `ROM = 2` ADD/JAL root-soundness regression with a committed successor lookup.

## Verification
- `lake build`
- `trust/scripts/check-all.sh`
- `trust/scripts/check-all-semantic.sh`
- `nix run .#test`

Reviewer pass: no blockers.

### DIFF
--- a/ZiskFv/Compliance/AcceptedZiskTrace.lean
+++ b/ZiskFv/Compliance/AcceptedZiskTrace.lean
@@ -6,7 +6,8 @@ import ZiskFv.AirsClean.FullEnsemble.Balance.TableProjections
 # Accepted trace
 
 An `AcceptedZiskTrace` is one fully-populated, verifier-accepted run of the RV64IM
-circuit for a fixed `program` of `length` instructions.
+circuit for a fixed committed `program` and an executed trace. Execution can revisit
+committed instructions, so its step count need not equal the ROM length.
 
 The circuit is an **Air.Flat ensemble**: a collection of AIRs (the Main table
 plus the binary / arithmetic / memory / lookup tables) wired together by shared
@@ -56,11 +57,11 @@ def acceptedMemReplayFixedSegment
     discharge it by proving their mutable-Mem table is empty, while memory-carrying traces provide
     the same bridge package #115 already required. -/
 def MutableMemPresent
-    {numInstructions : Nat}
-    {program : ZiskFv.AirsClean.ZiskInstructionRom.Program numInstructions}
+    {programLength : Nat}
+    {program : ZiskFv.AirsClean.ZiskInstructionRom.Program programLength}
     (witness :
       Air.Flat.EnsembleWitness
-        (ZiskFv.AirsClean.FullEnsemble.fullRv64imEnsemble numInstructions program).ensemble) :
+        (ZiskFv.AirsClean.FullEnsemble.fullRv64imEnsemble programLength program).ensemble) :
     Prop :=
   ∃ table ∈ witness.allTables,
     table.component = ZiskFv.AirsClean.Mem.componentWithDualMemBus ∧
@@ -76,7 +77,8 @@ def MutableMemPresent
     `2 + 4*i` / `3 + 4*i` in their natural Goldilocks representatives. -/
 structure MainStepIndexFixedFacts
     (numInstructions : Nat)
-    (program : ZiskFv.AirsClean.ZiskInstructionRom.Program numInstructions)
+    (programLength : Nat)
+    (program : ZiskFv.AirsClean.ZiskInstructionRom.Program programLength)
     (table : Air.Flat.Table FGL) : Prop where
   main_step_eq_index : ∀ i : Fin numInstructions,
     (ZiskFv.AirsClean.FullEnsemble.mainTableRowAtOrZero program table i.val).rom.main_step =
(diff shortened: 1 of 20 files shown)


codygunton:
## Reviewer pass — approve

The split is done in the honest direction at every seam checked, the protected-surface constraints from the Stream 2 spec are met, and V1 was independently re-run green (exit 0) at the PR head in a clean worktree.

### Soundness-critical calls — all made correctly

- **`AcceptedZiskTrace numInstructions` keeps executed steps as the structure parameter** (shared with `SailTrace`/`root_soundness`) and adds `programLength` as a dependent field with `program : Program programLength`. Deliberately **no relation** between the two — right, since spin loops execute more steps than ROM entries and padded ROMs are longer than executions. `main_height` still ties executed indices to physical rows.
- **Decode disambiguation strengthened, not weakened:** every `ProgramDecode_<op>.h_prog` (all 63 ops) now quantifies over `Fin trace.programLength`. This is the direction that matters — an executed-only quantifier would leave decode ambiguous when a longer ROM contains a second entry at the same pc. Likewise `mainRomMessage_at_eq_program`'s existential ranges over the full committed ROM and holds for **every physical row** including padding rows, so the lookup was not weakened to hide last-row issues.
- **Main's component is parameterized by `programLength`** (`componentWithRomMemAndOpBus trace.programLength trace.program` throughout), keeping the in-circuit ROM lookup committed-ROM indexed.
- **`MainStepIndexFixedFacts numInstructions programLength …`** correctly keeps the timestamp pins scoped to executed indices only.

### The regression is the right artifact

`addPaddedAcceptedTrace : AcceptedZiskTrace 1` with `programLength := 2` reuses the 2-row AddSpin witness: one executed ADD, a committed ADD/JAL ROM, `root_soundness` applied (`addPaddedRootSoundness`), plus `addPaddedSuccessorRow_hasCommittedLookup` / `_isCommittedJal` explicitly demonstrating the physical successor row stays committed-ROM-bound. Its ADD `h_prog` genuinely case-splits over both ROM entries. All five new endpoints are wired into the `#print axioms` gate hook.

### Hygiene

- No new `axiom` / `opaque` / `sorry` / `native_decide` / `partial` / `@[extern]` anywhere in the diff (grepped); V1's zero-sorry, locality, floors, and baseline-freshness checks pass at the PR head.
- The two `trust/generated/` baselines changed timestamp only (0 axioms before and after). `trusted-base.md` prose updated honestly — the new `h_idx` paragraph correctly reframes the successor-row condition as physical-row existence, separate from ROM length, with the #103 boundary preserved.
- All four pre-existing witnesses (degenerate, single-ADD, ADD-spin, #250's AddAddiSpin) rebased in the same PR, as the stream spec requires.
- `OpBusProviderMatch.lean`, `RomDecodeBindingOps.lean`, `BootSegmentMemorySeed.lean` are purely mechanical re-parameterization — every changed line in those files was filtered and none touches anything beyond the `numInstructions → programLength` substitution.

### Caveat before merge

No CI checks ran on the branch, and PRs are not V2-gated in CI. The PR body claims local `lake build` + V2 + `nix run .#test`; V2 was not cheaply re-verifiable during this review. Given main has gone red exactly this way once before, confirm (or re-run) `nix run .#test` on the final rebase before merging.

Optional nit, not blocking: `addPaddedBootSeed` / `addPaddedRowsOf_empty_readSound` duplicate the AddSpin empty-seed boilerplate almost verbatim — a shared empty-seed builder would trim ~30 lines across the four witnesses; cleanup for a later pass.

🤖 Review by [Claude Code](https://claude.com/claude-code)

PR #255, opened 2026-07-15
Make Lake dependencies Aristotle-compatible (pure-Git dependency graph)

codygunton:
## What

Makes the Lake dependency graph pure-Git so the project builds from a bare clone with only elan — no `nix run .#populate` before `lake build`:

- The three path dependencies (`build/sail-lean` → LeanRV, `build/aeneas-lean` → patched Aeneas runtime, `build/clean-lean` → Clean) are replaced by two immutable Git packages: **`ZiskFvLeanInputs`** (`eth-act/zisk-fv-lean-inputs @ 087a0de2`, the Nix-generated Sail RV64D tree + patched Aeneas runtime as one snapshot) and the **Clean fork pinned directly by rev**.
- Nix is demoted from build-time dependency to **generator + verifier**: `nix build .#aristotle-inputs` reproduces the snapshot from the flake-pinned upstreams, and `nix run .#check-aristotle-inputs` byte-compares the published package against that reproduction, rejects any root path dependency, and rejects a snapshot-root manifest. `scripts/sync-aristotle-inputs.sh` publishes a regenerated snapshot (guarded against overwriting the source checkout or an unrelated repo).
- `clean-src` is removed from the flake inputs; `nix/clean.nix` is deleted. **`lakefile.toml` is now the single pin for Clean** (previously the rev lived in `flake.lock` and reached Lake via a populate-copied tree).
- Second commit: syncs that Lake Clean pin to `c87617d8` — mirroring the bump #254 made on main's flake side — so the rebased graph builds the same Clean the proofs on main now require.

## Why

The immediate trigger was Aristotle compatibility (its Lake client cannot resolve path dependencies or pre-populated build trees), but the change stands on build-structure grounds alone: it removes the populate-before-build footgun and the stale-`build/`/symlink failure class, makes every dependency immutable and content-addressed, and turns the generated-inputs provenance into a public, auditable artifact backed by a CI byte-compare instead of a locally materialized tree.

`build/extraction` (PIL extraction) remains outside the Lake graph, exactly as before. No proof statement, trust ledger entry, or generated artifact changes.

## Verification

- `nix run .#check-aristotle-inputs` — snapshot byte-match + no path deps ✓
- `nix flake check --no-build`, `bash -n` on both new scripts ✓
- Full `lake build` + V1/V2 trust gates run green at the tip of this stack (see stacked PRs); this PR's Lean surface differs from main by one comment line.
- Pre-rebase, the original branch passed the full gate suite including the split Aeneas rerun (see branch history).

Part 1/3 of the Aristotle refactor stack (next: plan docs, then Phase 0+1.3).


codygunton:

PR #256, opened 2026-07-15
Check in the Aristotle architecture review and refactor plan

codygunton:
## What

Checks in the Aristotle architecture review and refactor plan under `docs/refactor/`:

- `FINAL-PLAN.md` — the folded, self-contained plan (root-theorem API redesign, Clean-as-spine, per-shape factoring, T0/T1/T2 root-stability discipline, upstream-Clean tracking, sequenced roadmap).
- `01`–`07` + `README.md` — the detailed backing documents it cross-references.
- Pointers from the top-level `README.md` and the workstream index.

Docs only — no Lean, trust, or build changes. The Phase 0 + 1.3 implementation PR sits on top of this one and its README fixes cite these documents by section (e.g. `05` C4), which is why the plan lands in-tree rather than staying local.

## Review notes

The plan's factual claims were spot-checked against the tree and hold (e.g. `Valid_Main` in exactly 299 files; Clean's `Vm`/`OrderedChannel` ensemble soundness used in 0 files). One known correction to carry into Phase 1 planning: §4.1's proposal to lift `aeneasBridgeTrust` / `memoryTimelineConstructionEvidence` into root binders is based on a stale reading — those envelope fields are already discharged internally by `StepStrong*` (from accepted-trace row facts / the `bootSeed` binder), so lifting them would weaken `root_soundness`, not clarify it. The plan's own T2 gating already marks that step as separate-and-deferrable; it should simply be dropped.

Part 2/3 of the Aristotle refactor stack (base: the Lake portability PR; on top: Phase 0+1.3).


codygunton:

PR #276, opened 2026-07-18
Phase 3: Airs/Arith dissolved; live supply swapped to mainComplete (T11)

codygunton:
Phase 3 of the architecture refactor (`docs/refactor/FINAL-PLAN.md`), turn T11: **`ZiskFv/Airs/Arith/` is gone and the live Arith supply now runs through the completed `mainComplete` circuits.** This turn had an unusual operator loop — details below, because the process matters for auditing it.

## How this turn was produced

1. Aristotle's first run (task `3411ddf3`) was deliberately paused at 1h31m while it ground the projection lemmas; its committed scaffolding (named spec bundles) was received, and I filled the two projection-theorem sorries locally (`complete_local_specs_of_const_soundness`, 19 + 49 clauses, both first-try green) — then resubmitted with item 1 marked done.
2. The resumed run (task `4af9cac5`) was cancelled at 6h15m on a stall diagnosis that turned out to be **wrong** — the progress feed was misleading, and its committed state actually contained the full migration. Everything here from that state was received, deletion-replayed, and verified locally from scratch.

## What changed

- **Live-supply swap (the substantive step):** `ArithMulTableWitness` / `ArithDivTableWitness` in `Compliance/SharedBundles.lean` now carry `ConstraintsHold.Soundness` of **`mainComplete`**. The witness *constructors* prove the appended selector constraints from static Arith-table membership (`mode_spec_of_arith_table` — every table row has boolean/disjoint selectors, proven by `fin_cases` over the table), which is precisely the constructibility argument for the strengthening. Witness *consumers* keep their old projections via `base_soundness_of_complete_const_soundness` (forget the appended constraints, retain the base circuit supply).
- **`Airs/Arith/` dissolved (5 files):** `CarryChain`/`CarryChainCompleteness` → `AirsClean/ArithShared/` (byte-identical moves), `BusRes1` → same (4 lines), and the `Valid_ArithMul`/`Valid_ArithDiv` record models → `AirsClean/Arith{Mul,Div}/Semantics.lean` (2 changed lines each), **keeping their original namespaces**. The namespace retention is forced, not lazy: the frozen DIV/REM defect boundary (`Compliance/Defects.lean`, byte-unchanged as required) is typed over `Valid_ArithDiv`, so the record must keep existing under its old name until the defect boundary itself is ever re-scoped (a human-gated decision).
- ~50 consumer files repointed via 2-line import retargets.
- Projection layer complete: named spec bundles + proven projections for both families; no consumer destructures `mainComplete`'s raw conjunction.

## What this turn is NOT

The record-model constraint predicates (`mul_mode_booleans` etc.) are still consumed as hypotheses in places where they should eventually be *derived* from the Clean supply — that derivation sweep plus the final reference audit is the next turn's work order. This PR is the file-level dissolution + supply swap, honestly labeled.

## Review performed (operator)

- Zero diff lines touch `state_effect_via_channels` / `execute_instruction` / `bus_effect`; zero added trust markers; `trust/generated/`, roots, `Audit.lean`, pins, `Defects.lean` byte-unchanged.
- All five deleted files' contents accounted for as verified moves (0/0/4/2/2 changed lines).
- SharedBundles' witness-swap read in full — the strengthened witness is constructible (table membership ⇒ new constraints) and consumers are projection-compatible.
- Zero remaining references to `Airs.Arith` paths.

## Verification (local, this machine)

- `lake build` — full, green (9,015 jobs).
- `trust/scripts/check-all.sh` — 16/16 including check 13.
- `trust/scripts/check-all-semantic.sh` — 16/16.


codygunton:

Issue #279, opened 2026-07-19
Live ensemble validates an incomplete Arith provider circuit — model-coverage gap class


codygunton:
## Finding (verified 2026-07-19, refactor turn R14; full audit in `REFACTOR_14_REPORT.md` on branch `refactor-14`, PR #278)

The accepted-trace ensemble validates arithmetic provider rows through the shared `ArithMul.componentWithArithTable` component, whose circuit (`mainWithArithTable`) mirrors only a **subset** of the physical Arith AIR's generated constraints — the carry chains, table/range lookups, and (as of the refactor stack) the selector booleans. The Div-block generated constraints (zero-divisor/overflow boundary `constraint_9–24`, inverse-sum detector `constraint_25`, scope/disjointness `constraint_26–30`, W-mode lanes `constraint_47–48`) are **not validated by the live ensemble**:

- `ArithMulRow` does not model the Arith AIR's `inv_sum_all_bs` witness column (stage-1 col 38), and the row-view converter `vOfDivuRow` fabricates it as `0` — under which the generated inverse-sum equation is *false* on ordinary nonzero DIV/REM rows.
- The complete constraint mirrors exist in Lean (`AirsClean/ArithCompleteConstraints.lean`, every `assertZero` cited to its `constraint_N_every_row` extraction fact + `arith.pil` line) but as **dormant** circuits the ensemble never runs.
- Consequently the Div-family wrappers carry 16 caller-supplied row-constraint premises that cannot currently be derived. An attempted derivation (refactor turn T13/R14) was correctly refused as unconstructible — discharging them against the incomplete live supply would have been obligation laundering.

## Why this matters beyond Arith

This is a **class**, not a one-off: "the Clean circuit the ensemble actually validates" can silently cover fewer generated constraints than the Lean mirrors or the legacy `Valid_<AIR>` models suggest. Wherever that happens, facts get carried as caller premises instead of derived, and the modeled acceptance predicate is weaker than the real AIR. The per-family Q2 audits compare constraint *lists*; they do not check *which circuit the ensemble validates* — that dimension needs its own audit (in progress: a live-provider coverage matrix across Main/Mem/MemAlign/RegisterBoundary and the static-table families).

## Remediation

[...]

Issue #282, opened 2026-07-19
Five vestigial _of_table compatibility binders pinned by frozen Equivalence/ signatures (owner decision)


codygunton:
Found during refactor turn R13/T12 (PR #277) and parked as owner-gated; filing so the decision isn't lost in PR bodies.

## Finding

After the Mul-family wrapper discharge, five `_of_table` declarations retain ignored positional compatibility binders. They are semantically dead (the facts are now derived from the Clean supply via `ArithMulTableWitness.row_constraints`), but removing the parameters would change the signatures of byte-frozen `Equivalence/` public theorems — a protected-interface change requiring explicit owner approval per AGENTS.md.

## Options

1. **Wait for Phase 4** (recommended in-thread): the planned envelope/dispatch collapse reworks exactly these surfaces; the binders die there for free.
2. **Dedicated signature-cleanup PR now**: small, but touches the protected `Equivalence/` surface and invites churn Phase 4 would redo.

The Div-side analogues (16 obligations) are separate and tracked via #279's remediation (refactor turn R15): they are real obligations until the live provider completes, not vestigial binders.

codygunton:
Closing as moot: the compatibility binders this issue describes existed only on the Aristotle refactor stack (turn R13, PR #277), which was closed unmerged per #323 — they were never on main. No action remains.

PR #283, opened 2026-07-19
Phase 3: live Arith provider completed; ensemble swapped to componentComplete (R15)

codygunton:
Phase 3 turn R15 — the remediation for #279, and the load-bearing move of the whole Arith arc: **the live ensemble now validates the completed Arith provider circuit.**

Authored by Aristotle (task `f1bb6e7f`), received, decontaminated, and verified locally.

## What changed

- **Constructibility precheck first (item 1, committed before code):** every shared provider row family audited against every constraint to be added — **no unsatisfiable row**. Active nonzero DIV/REM rows require the actual inverse of the divisor-chunk sum; multiplication and division-by-zero rows legitimately use zero.
- **`inv_sum_all_bs`** (the physical Arith AIR's stage-1 col 38, which `ArithMulRow` never modeled — the root cause found in R14/#279) added with mechanical propagation.
- **`componentComplete`**: the shared circuit extended with the full frozen generated local-constraint set (every `assertZero` cited), proven component-sound.
- **The swap**: ensemble, balance layer, six construction modules, and fixtures moved uniformly from `componentWithArithTable` to `componentComplete`; `vOfDivuRow` projects the committed inverse witness instead of fabricating `0`.
- Net **+161** proof-code lines (the completed circuit + soundness; deletions resume with the wrapper discharge).

## Reviewer note: what the swap means

The modeled acceptance predicate is now a strictly more faithful mirror of the real AIR (it validates constraints the physical AIR has always enforced). Constructibility is proven by the repaired witnesses, per the item-1 precheck. Root statements, `Audit.lean`, `Defects.lean`, `Equivalence/`, and `trust/generated/` are byte-unchanged; V1 16/16 (check 13 run locally with the submodule) + V2 green.

## Honest residual

The ArithDiv-view witness upgrade and the 16 Div wrapper discharges remain blocked on one precisely-named seam (`RowExtraction.arithMul_fullSpec_of_component_spec` deliberately keeps the six-part `FullSpec` ABI; the complete facts live in the table's `constraints_hold` but aren't yet transported through the row view). Per the ROI plan, that seam is being closed **locally** by the operator rather than with another Aristotle turn; it lands as the next commit on this stack.

### DIFF
--- a/ARISTOTLE_SUMMARY.md
+++ b/ARISTOTLE_SUMMARY.md
@@ -1,3 +1,40 @@
+# Summary of changes for run f1bb6e7f-eec5-4d3b-8a81-c9e1e49bee44
+Completed Refactor 15 through the verified live-provider boundary and documented the result in `REFACTOR_15_REPORT.md`; prepended the required summary to `ARISTOTLE_SUMMARY.md`.
+
+Key results:
+- Installed and byte-verified all 840 archived files/symlinks non-destructively, preserving `.lake`; attached tarballs were not committed.
+- Audited every shared Arith provider row family. No unsatisfiable row was found. Active nonzero DIV/REM rows require the actual inverse of the divisor-chunk sum; multiplication and division-by-zero rows may use zero.
+- Added and mechanically propagated the physical `inv_sum_all_bs` witness column.
+- Added the full frozen generated local-constraint set to a completed shared Arith circuit and proved its component sound.
+- Swapped the live ensemble, construction modules, and fixtures to the completed provider.
+- Updated `vOfDivuRow` to project the committed inverse witness rather than fabricate zero.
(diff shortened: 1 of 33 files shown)


codygunton:
Follow-up commit on this branch closes the residual seam noted in the PR body: the Div-view projection is in, `ArithDivTableWitness` carries the completed circuit, and the eight Div wrappers' caller-supplied row bundles are now derived from the live supply (frozen-callable surfaces keep documented ignored compat binders, mirroring the Mul side). Implemented locally per the derisk plan — no additional Aristotle turn. Full build + V1 (16/16 incl. check 13) + V2 green on the committed tree. The Arith family is closed except the boundary-class obligations, whose facts are confirmed available in the same projection (fast follow-up), and the owner-gated items (#282, defect-boundary re-scoping).

codygunton:

PR #284, opened 2026-07-19
Phase 4 pilot turn 1: R-type provider elimination; class collapse blocked (R16)

codygunton:
**Pilot turn 1 of 2 for the go/no-go experiment** (budget: ≥5,000 net lines deleted in ≤2 turns, else the refactor stops — owner ruling). Authored by Aristotle (task `dcea4ffd`, COMPLETE_WITH_ERRORS), received and verified by the operator.

## Verdict data, plainly

- **Net delta: +49 lines. Zero per-opcode files deleted.** The ≥5K target was not approached.
- Landed and real: the full 22-opcode factoring map (in `REFACTOR_16_REPORT.md`); `Compliance/ParametricStaticBinary.lean` — two *proved* provider-elimination theorems (`dischargeRType`, `discharge`) abstracting the repeated Clean static-Binary witness projection; AND/OR/XOR wrappers migrated onto it (−10 repeated lines each).
- **The structural blocker (verified, well-stated):** a per-class exact-conclusion parametric theorem needs a dependent packaging interface relating each opcode's *generated Sail* input structure, execute function, and instruction constructor to the bus-effect normal form. The existing `WriteValueProofs` consolidation abstracts only the semantic core. Compare and shift classes add signed/carry and width/sign-extension pins on top of that. This interface is precisely FINAL-PLAN Phase 4.1's "per-shape envelope" — the pilot established it must exist *before* any class collapses, not after.
- **Performance red flag:** direct generic migration of ANDI/ORI/XORI hit the 1,200,000-heartbeat elaboration ceiling (edits reverted, recorded with file:line). Dependent abstraction over generated types is elaboration-expensive — a material risk to the whole 63-opcode collapse thesis even once the interface exists.

## Operator review record (per the PR policy)

- Independent verification: full `lake build` (9,016 jobs), V1 16/16 (check 13 run locally with the submodule), V2 16/16.
- Diff read in full (small: 4 proof files + report): the two parametric theorems are genuinely proved (no trust markers), the wrapper migrations are strict simplifications, and the reverted experiments left no residue. `Equivalence/` statements byte-identical; all frozen surfaces untouched.
- Steering record: none required this turn — the honest-partial shape and blocker documentation followed the work order's contract as written.

## What this means for the go/no-go

Turn 2, as originally scoped, would have to design + build the dependent interface, instance it for the class, prove the parametric theorem, migrate, and delete — in one turn, against a demonstrated elaboration-performance hazard. The operator's assessment and options are with the owner; this PR is the evidence record.

### DIFF
--- a/ARISTOTLE_SUMMARY.md
+++ b/ARISTOTLE_SUMMARY.md
@@ -1,3 +1,30 @@
+# Summary of changes for run dcea4ffd-3717-42e9-b5ba-6fab6f389dc0
+Completed and documented the Refactor 16 pilot.
+
+- Installed the authoritative archive non-destructively and byte-verified all 842 archived files/symlinks while preserving `.lake`.
+- Produced the full 22-opcode shape inventory and factoring map in `REFACTOR_16_REPORT.md` before migration.
+- Added a proved reusable static-Binary provider-elimination layer and migrated the AND/OR/XOR R-type wrappers to it.
+- Recorded precise blockers for the remaining logic, compare, and shift collapse, including the tested I-type elaboration failures and missing dependent opcode interface.
+- Reported the exact retained proof-code delta: +97/−48, net +49 lines. The ≥5,000 deletion target was not met, and no per-opcode tower files reached zero references.
+- Prepended the run summary to `ARISTOTLE_SUMMARY.md`.
+
+Verification completed:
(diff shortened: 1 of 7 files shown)


codygunton:

PR #300, opened 2026-07-28
Weld the Arith constraint mirror to the generated extraction

### LINKED ISSUE #296: Weld the hand-written Arith constraint mirror to the generated extraction (drift is currently unchecked)
Found by the PR #294 review (S8 / #279), now merged to `main` as `2da97416`.

## Finding

`ZiskFv/AirsClean/ArithCompleteConstraints.lean` is a **hand-written mirror** of the generated Arith
constraints, and nothing in the build checks that it still matches. Its own header says so:

> Each group below is a handwritten circuit mirror of the named generated
> `Extraction.Arith.constraint_N_every_row` definitions. The comments are source
> citations, not machine-checked links; equivalence was audited against the current
> generated extraction.

So the citations (`-- constraint_25_every_row; arith/pil/arith.pil:92`) are **comments**. A
transcription slip — `65535` for `65536`, a dropped `(1 - m32)` factor, a mis-parsed
`1 - main_mul - main_div` — compiles fine and silently proves soundness of a circuit that is not
ZisK's. The same is true of the pre-existing mirrors this file extends (`ArithMul/Constraints.lean`,
`Valid_ArithDiv` in `Airs/Arith/Div.lean`, etc.).

This is not a claim that the current mirror is wrong. PR #294's review audited all 41 newly asserted
identities in `sharedMainComplete` line-by-line against `build/extraction/Extraction/Arith.lean` and
found **exact** matches, including preserved generated redundancies (c29 `overflow*div0` and c30
`div0*overflow` both present) and exact association in c25's inverse-sum. The gap is that nothing
*keeps* it exact: `nix run .#populate` regenerating a changed upstream PIL would not fail any gate.

## Why now

#294 substantially enlarged this hand-mirrored surface (the whole Arith Div block: `constraint_9-30`,
`47-48`, the `inv_sum_all_bs` column, and the remainder-bound bus emission) and swapped the live
ensemble onto it (`componentComplete`). The mirror is now load-bearing for signed-DIV soundness and
for the concrete DIV `root_soundness` instantiation.

## Fix shape (spike first — feasibility is the open question)

Weld the mirror to the generated definitions so drift is a build error, not a silent divergence.
Precedent: #160 welded `ExtractedConst` ↔ `ZiskOp.code` for decode constants; #111/#160's
`trust-gate check-extraction-closure` already auto-gates the decode side.

The obstacle to price honestly in the spike: the two sides are stated over **different
representations**. Generated constraints are untyped column accesses over the `Extraction.Circuit`
class —

```
Extraction.Circuit.main c (id := 1) (column := 36) (row := row) (rotation := 0)
```

— while the mirror is structured field access on a `Var ArithMulRow FGL` (`row.flags.div_by_zero`,
`row.chunks.b_0`). A weld therefore needs an explicit **column-index ↔ field** mapping plus a proof
that each `assertZero` argument is the same polynomial as its `constraint_N_every_row` under that
mapping.

Spike deliverables:

1. Decide the weld's shape: per-constraint `rfl`/`decide` against an instantiated `Extraction.Circuit`
   instance backed by the structured row, vs. a generated-side evaluator, vs. a checked
   column-mapping table plus per-constraint lemmas.
2. Prove it for **at least three representative constraints** of different shapes — a plain boolean
   (`constraint_39`), a boundary pin with an `m32` factor (`constraint_15`), and the inverse-sum
   detector (`constraint_25`) — before committing to the full set.
3. Report the real cost of the remaining constraints and whether it generalizes to the other AIR
   mirrors (Main, Binary, Mem), or state plainly that it does not.

**Anti-laundering constraint on this issue:** a weld that is vacuous, or that is proved by
restating the mirror as the definition of the generated side, is worse than no weld — it would
convert an honest, documented hand-audit into a false machine guarantee. If the spike concludes the
weld cannot be made real, say so and close with that finding recorded; do not ship a trivial
`rfl`-shaped tautology.

## Acceptance

- Either: a build-enforced equivalence covering `sharedMainComplete`'s asserted constraints, whose
  failure mode on a deliberately mutated mirror line is a **build error** (demonstrate this with a
  temporary mutation in the PR description);
- Or: a written spike verdict explaining precisely why it is not achievable, with the residual
  recorded in `trust/trusted-base.md` as a named hand-audit dependency rather than left implicit.
- No new `axiom` / `sorry` / `native_decide`; `lake build` + both trust suites green.

Related: #279 (source of the enlarged surface), #160 (decode-constant weld precedent), #184 (the
premise audit that should inherit an accurate statement of what is machine-checked vs. hand-audited).

codygunton:
Closes #296. Stream S8b.

Welds the hand-written Arith constraint mirror to the generated extraction, so drift between them becomes a build error instead of a silent divergence.

## Why

`ZiskFv/AirsClean/ArithCompleteConstraints.lean` is a hand transcription of the generated Arith constraints, and its own header admits the gap: *"the comments are source citations, not machine-checked links."* A slipped `65535`, a dropped `(1 - m32)` factor, or a mis-parsed mux compiles fine and silently proves soundness of a circuit that is not ZisK's. #279 substantially enlarged that surface and swapped the **live ensemble** onto it, so the mirror is now load-bearing for signed-DIV soundness.

## What this does

Constraints of deliberately different shapes are welded to `Extraction.Arith.constraint_N_every_row` as compiled **`Iff.rfl`** theorems — a definitional tie, not a restatement:

- `constraint_39` — plain boolean
- `constraint_15` — boundary pin with an `m32`-weighted factor
- `constraint_25` — the inverse-sum detector
- `constraint_36` — the one member of the F-only set that reads a carry lane

Plus `trust/scripts/check-arith-column-map.py` as V1 check **19/19**, pinning the 44-entry stage-1 column map.

## Non-vacuity, demonstrated

Mutating a welded literal breaks the build. Changing column `a_2` → `a_3` in `constraint_15_weld`:

```
error: ArithMirrorWeld.lean:160:2: Type mismatch
  Iff.rfl has type ?m ↔ ?m
  but is expected to have type
    row.flags.div_by_zero * (row.chunks.a_3 - (1 - row.flags.m32) * 65535) = 0 ↔
      Arith.extraction.constraint_15_every_row (extractedArithRow row) …
```

Mutating a column-map arm breaks the gate. Both reverted; tree clean.

## The fail-open hole, and why the obvious fix was wrong

Review of the spike found the column-map check returned 0 with a SKIP when `build/extraction/Extraction/Arith.lean` was absent. That is worse than it sounds: `.github/workflows/trust-gate.yml` checks out without `nix run .#populate`, so **the V1 gate on every push to main would have taken the SKIP branch 100% of the time and checked nothing.**

Hard-failing on absence would turn that workflow red. So the check's declared input is now *tracked* — `trust/generated/arith-stage1-columns.txt`, written by `regenerate-arith-stage1-columns.py`, which imports the gate module and reuses **its** parser so recording and gate cannot disagree. Missing/empty recording, missing weld, zero arms, or a generated file with no `stage 1 col` header now all exit 1 with `refusing to pass vacuously`. When the generated file *is* present, the recording must reproduce it exactly. This mirrors the existing `check-aeneas-generated-bridge-manifest` idiom.

## Regex hardening

The arm regex captured only the last path segment, so a sub-record swap that kept the field name passed. It now resolves both segments against the real `ArithMulRow`/`Chunks`/`Flags`/`Carries` declarations parsed from `Row.lean` (comment-stripped with the same nestable-block-comment stripper `check-module-reachability.py` uses), rejects unknown sub-records/fields and duplicate cells, and **checks its own premise** that no field name occurs in two sub-records rather than assuming it.

## Scope, stated plainly

**All 49 of the Arith AIR's F-only constraints are now welded** (the PR originally welded 4; the remaining 45 landed in `13d723fb` via five predicate-level `Iff.rfl`s plus `spec_carryChain_weld`). `extracted_of_sharedMainComplete` grows from 3 to 37 generated constraints derived from the live `sharedMainComplete` assertion list, with its hypothesis list unchanged — only its conclusion strengthened.

Generated constraints 49–64 are outside the weld entirely: they read stage-2 lanes and are not part of the F-only slice.

Honest limits, stated in the module rather than buried:
- 12 constraints (6–8, 31–38, 46) are welded against `Spec`/`C46Spec` as *published predicates*; this module does not re-derive those two from the component's own `assertZero` list.
- `spec_carryChain_weld` is an implication, so unlike the five `Iff` welds it does not forbid `Spec` asserting strictly more than the AIR does.
- The probe bridge is instantiated at four representative constraints; the lane-restriction result it composes with is proved for all 49.
- `restrictToModeledLanes`'s `column < 44` bound is sound but not tight — the F-only set reads at most column 40.

## The binding hazard — closed with a Lean-level tie, and demonstrated

The earlier version of this PR disclosed a hazard: the gate pinned `mainValue`'s 44 arms but never checked that `mainValue` is what the circuit instance's `main` field is **bound** to.

Closed by `extractedArithRowCircuit_pinned`, stated through `inferInstance` (so it also catches a *shadowing* instance) and placed last in the module so every weld above resolves against instances it can see.

Demonstrated rather than asserted — rebinding `main` at **column 41**, which no F-only constraint reads, so no existing weld notices:

```
error: ArithMirrorWeld.lean:473:0: Not a definitional equality: the left-hand side
  inferInstance is not definitionally equal to the right-hand side
  { main := mainValue, preprocessed := …, challenge := …, exposed := … }
```

**Control on the same mutated tree:** `check-arith-column-map.py` printed `44 stage-1 columns pinned` and exited 0. So the script alone would have missed it; the new theorem is the only thing that catches it.

## A second hole of the same shape, found and fixed

While closing the above, the agent found `check-arith-column-map.py` collected arms from *anywhere* in the file, so arms parked in a helper that `mainValue` delegates to would be pinned by the gate but unpinned by the Lean theorem (which names `mainValue`). It now requires exactly one top-level `def mainValue` and reads arms only from that body. A defect in *that* fix — a `\b` boundary accepting `mainValue'` — was caught by its own mutation test and fixed.

## Stub lanes: answered with a proof, not prose

`mainValue` returns `0` for `id ≠ 1` and out-of-range columns, and `preprocessed`/`challenge`/`exposed` are stubs. Rather than assert those are unobservable, `ArithProbe` is the **free** `Extraction.Circuit` (every lane an unconstrained function field), and `fOnlyConstraints_readOnlyModeledLanes` proves by `Iff.rfl`, for **all 49** F-only constraints, that restricting a probe to stage 1 / columns < 44 / rotation 0 leaves them unchanged.

Positive control: asserting the same property of `constraint_64_every_row` (which does read `preprocessed`/`exposed`) fails to elaborate — so the property is not vacuous.

## Gates

- `lake build` — green, 9120 jobs; `ArithMirrorWeld` is in the **default** build graph (not an explicit `+module` target)
- `check-all.sh` — **19/19**, numbering verified programmatically as contiguous 1..19, no gaps or duplicates; main's 18/18 reachability preserved verbatim
- `check-all-semantic.sh` — 19/19, unchanged

No new axioms/`sorry`/`native_decide`; all weld theorems close over `{propext, Classical.choice, Quot.sound}`.


codygunton (commented):
Reviewed at 641c6bcc. No blocking findings. The 49 F-only welds, live-assertion links, lane-restriction probes, instance pin, and source/generated column recording are internally consistent; focused Lean build passed. Approved for the requested merge sequence, with #310 to replace the Arith-specific gate after the full fan-out lands.

Issue #159, opened 2026-06-27
RomImageBinding: bind the in-build Aeneas decode pins to the committed program (make #111 load-bearing in root_soundness)


codygunton:
> **Child of #61.** This is the dedicated, decomposed record of the "real Aeneas / ROM binding"
> premise that #61's success criteria name but do not break down. #61 stays the umbrella for deriving
> *all* per-row witnesses; this issue is specifically the program/decode-fidelity pillar that makes
> #111's in-build decode pins load-bearing.

## Summary

#111 discharged all 63 RV64IM **static decode pins** in-build, kernel-soundly, from the real
Aeneas-extracted lowerer (`ZiskFv/Compliance/AeneasBridgeTrust/Extraction/`, axioms
`{propext, Classical.choice, Quot.sound}`). But those theorems are **standalone** — they prove
*"the real ZisK lowering of a raw word `W` produces these decode fields"*, and nothing yet ties them
into `root_soundness`. This issue tracks **RomImageBinding**: making the decode pins *load-bearing*,
i.e. turning `root_soundness`'s currently-**assumed** decode facts into **derived** ones.

## Why it isn't already load-bearing

Three facts about the current soundness statement (all verified):

1. **`root_soundness` never mentions a raw 32-bit instruction word.** `AcceptedZiskTrace.program` is
   `Fin n → ZiskRomMessage` — a sequence of *already-decoded* 11-field messages (op, immediates,
   offsets, flags), not bytes (`ZiskFv/Compliance/AcceptedZiskTrace.lean`,
   `ZiskFv/Channels/ZiskRomBus.lean`).
2. **The decode facts are caller-assumed, not derived.** `root_soundness` takes
   `rowDecodes : ∀ i, RowDecode …`; each `Decode_<op>` (e.g. `Decode_add` in
   `TraceLevelExport/RowDataArithMem.lean`) asserts `mainOfTable … |>.op i = OP_ADD`, … as
   **hypotheses** — they are not proved from `AcceptedZiskTrace.constraints_hold`.
3. **The Sail side takes a pre-decoded instruction** (`execute (instruction.RTYPE (r2,r1,rd,.ADD))`),
   with no Sail decode of a raw word on the soundness path (the completeness side has one in
   `Completeness/Rv64im/SailDecode`, soundness does not).

The circuit *does* have a Rom AIR + a Main↔ROM lookup, and it *is* in the composed ensemble
[...]

Issue #111, opened 2026-06-19
Discharge aeneasBridgeTrust: connect the real Aeneas extraction in-build (replace the loose cross-workspace coupling)


codygunton:
## What

`OpEnvelope.aeneasBridgeTrust` (`ZiskFv/Compliance/AeneasBridgeTrust/Base.lean:19`) is the
per-arm decode/lowering bridge predicate: for each `OpEnvelope` arm it asserts the Main-row
decode/lowering facts ZisK's production lowerer would produce — `op`, `isExternalOp`, `m32`,
`setPc`, `storePc`, the row-mode, and the dynamic immediate/register/lane bridges. The predicate
is split under `ZiskFv/Compliance/AeneasBridgeTrust/` (root `AeneasBridgeTrust.lean` is the
aggregator; `Base.lean` holds the definition; `BinaryRType` / `Branches` / `ControlAndUType` /
`DivRem` / `ImmediateAlu` / `Loads` / `Mul` / `Shifts` / `Stores.lean` hold the per-family
builders).

It feeds two soundness theorems:

- **OpEnvelope level:** `zisk_riscv_compliant_program_bus` (`ZiskFv/Compliance.lean:95`) takes
  `h_bridge : env.aeneasBridgeTrust` (line 97) as an explicit premise.
- **Trace level (current headline):** `root_soundness` (`ZiskFv/Soundness.lean:36`) takes a
  per-row `rowDecodes : ∀ i, RowDecode …` bundle and dispatches through the `stepStrong_*`
  builders (`ZiskFv/Compliance/TraceLevelExport/StepStrong*.lean`), which construct
  `h_bridge : env.aeneasBridgeTrust` internally from those decode pins.

The bridge is **not an axiom** — the project closure is 0 project axioms. It is *discharged from
decode-pin hypotheses*: the per-op `OpEnvelope.aeneasBridgeTrust_<op>OfExtractedShape` theorems
(e.g. `…_luiOfExtractedShape`, `ControlAndUType.lean:42`) prove `aeneasBridgeTrust` for an envelope
built by `OpEnvelope.<op>OfExtractedShape` from hypotheses of the form
`provenance.extractedRow.op = ExtractedConst.opCopyB`, `.isExternalOp = false`, `.m32 = false`, …
(`MainExtractedRow` / `MainRowProvenance`, `ZiskFv/Compliance/RowProvenance.lean:48,123`).

[...]

Issue #61, opened 2026-06-06
Close the root_soundness rowData gap: decode-driven trace-level export


codygunton:
## Implementation complete; PR stack open (2026-07-30)

Phase 9's decode-driven export is implemented through the additive raw-program endpoint:

```lean
ZiskFv.Compliance.root_soundness_rawProgram
```

The implementation is split into this review stack:

1. #313 — production raw-word binding foundation
2. #314 — initial current-API raw program decoders
3. #315 — fidelity and non-vacuity hard gates
4. #316 — signed offsets and expanded JALR row layout
5. #317 — architectural decoder migrations
6. #318 — final control/JALR constructors and root endpoint

Review and merge in order. Each PR targets the preceding stack branch; GitHub should retarget the
next PR as its base lands. This issue remains open until all six PRs are merged.

## Result

The existing headline `ZiskFv.Compliance.root_soundness` and `AcceptedZiskTrace` are unchanged.
The additive theorem replaces the caller's
`programDecodes : ∀ i, ProgramDecode ...` binder with:

- a raw RV64IM image and caller-supplied architectural-to-physical layout
  (`rawProgram`, `start`, and `addr`);
- one exact `ProgramRowsBinding` relating each raw word to the rows emitted by the
  Aeneas-extracted production lowerer;
- per-step `RawProgramDecode` evidence tying the selected `ZiskStep` claim to that raw word and
  its required non-ROM witnesses.

`programDecode_of_rawProgramDecode` dispatches all 63 RV64IM families. JALR follows the real
production expansion: aligned instructions use one AND row, while unaligned instructions use
adjacent ADD+AND rows. Signed offsets and exact `rs1 = x0` / `rd = x0` selectors are preserved;
no alignment, nonnegative-offset, or nonzero-register premise was added.

## Explicit residual boundaries

This issue does not claim:

1. that the caller-supplied `start`/`addr` layout came from a particular linked binary;
2. the Sail-side `raw word → ext_decode` grounding tracked by #172;
3. that `rawProgram` is the intended compiled binary, which remains the compile/commitment
   boundary.

`inputsAgree`, `BootSegmentMemorySeed`, and `RowOutsideDefectRegion` remain the same named premises.

## Verification on the complete stack

- `lake build`: 9,132 jobs passed.
- `trust/scripts/check-all.sh`: 18/18 passed; 702 reachable modules and 28 semantic probes.
[...]

PR #301, opened 2026-07-28
Prove the ZisK lowerer succeeds for all 63 RV64IM opcodes (#61 S1)

codygunton:
Part of #61 (S9), spike **S1**. Design context: `docs/ai/research/NOTE_61_RAWWORD_JOIN_DESIGN.md`.

Proves the ZisK lowerer **succeeds** for all 63 RV64IM opcodes — a side condition that #111's static decode pins and the block-2 dynamic pins currently *assume*.

## Why this lands alone

#61's Phase 0 is blocked on the raw-word join. The design document deliberately scopes this spike to land independently: it has no dependence on the serialization the rest of route (c) must repair, it has the smallest merge surface, and it is a genuine trust reduction on its own — so it keeps its value even if the rest of the route dies.

## What is proved

Totality of the **real production lowerer** — `riscv2zisk_single_row.Riscv2ZiskContext.lower_rv64im_single_row_input`, reached through the Aeneas extraction in `ProductionM2.lean`, the same entry point the ELF transpiler uses. Not a re-implementation; a re-implementation would discharge nothing.

## Honest scope — three things a reviewer should not have to discover

1. **The 63 theorems are not premise-free.** `ri.rs1.val < 32`, `ri.rs2.val < 32`, `ri.rd.val < 32` remain caller-supplied (plus `ri.rom_address = 0#u64` for JALR, the transpile call site's value). The earlier sections prove those bounds for each *leaf* decoder, but they are not yet threaded into the per-opcode statements.
2. **Zero downstream consumers today.** The `_dispatch_static_pins` / `_dispatch_extracted_rowMode_pins` family is referenced nowhere else in the tree, so this discharges an obligation nothing currently invokes. That is the honest characterisation: it is a prerequisite banked, not a premise removed from a live theorem.
3. **~762 of the added lines are a verbatim recovery** of the unmerged June branch `a16bd97c` (#159 block 3), byte-identical where recovered, repaired where it had drifted against today's tree.

A premise-free `∃ e, extract_transpile_rv64im_raw raw = ok e` needs three further things beyond this spike, one of which is substantive rather than plumbing; they are named in the module.

## Gates

- `lake build` — green, 9119 jobs, **default target** (no `+module`), module imported from `ZiskFv.lean` as the reachability gate requires
- `check-all.sh` — 18/18
- `check-all-semantic.sh` — 19/19

No new axioms/`sorry`/`native_decide`.

### DIFF
--- a/ZiskFv/Compliance/AeneasBridgeTrust/Extraction.lean
+++ b/ZiskFv/Compliance/AeneasBridgeTrust/Extraction.lean
@@ -63,3 +63,4 @@ import ZiskFv.Compliance.AeneasBridgeTrust.Extraction.LoadStore
 import ZiskFv.Compliance.AeneasBridgeTrust.Extraction.Precompiled
 import ZiskFv.Compliance.AeneasBridgeTrust.Extraction.Dispatch
 import ZiskFv.Compliance.AeneasBridgeTrust.Extraction.DynamicFields
+import ZiskFv.Compliance.AeneasBridgeTrust.Extraction.Totality
--- a/ZiskFv/Compliance/AeneasBridgeTrust/Extraction/Totality.lean
+++ b/ZiskFv/Compliance/AeneasBridgeTrust/Extraction/Totality.lean
@@ -1,0 +1,1868 @@
+/-
+ZiskFv/Compliance/AeneasBridgeTrust/Extraction/Totality.lean  (eth-act/zisk-fv#61 / #111)
+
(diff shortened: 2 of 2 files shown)


codygunton:
Superseded by the phase9-* stack: these commits were rebased into #313 as c4482255/a748e5f0 (`Extraction/Totality.lean`, since extended). Closing in favour of the stack.

PR #302, opened 2026-07-28
Transcribe the ZisK ROM-row emitter and cross-check it (#61 S2)

codygunton:
Part of #61 (S9), spike **S2**. Design context: `docs/ai/research/NOTE_61_RAWWORD_JOIN_DESIGN.md` §4(c), §5.

A cited, line-by-line transcription of ZisK's ROM-row emitter (`zisk/state-machines/rom/src/rom.rs:204-260` plus the flag chain at `zisk/core/src/zisk_inst.rs:289-308`), written **twice in genuinely different shapes** and proven equal.

## Why the two-definition discipline

A raw-word binding (`trace.program k = <serialization of the lowering of word k>`) is only as strong as the serialization it names. A plausible-looking but wrong serialization yields green theorems over a false premise. The June asset at `a16bd97c` is the cautionary case: it is kernel-sound Lean and it is **wrong in three confirmed ways**, all verified here against the pinned Rust:

1. `SRC_IND` written as literal `3`; the truth is `5` (`3` is `SRC_STEP`). **Already falsified by a witness that ships in our semantic gate** — `SdLdSpinWitness` sets `b_src_ind := true`, which the June code computes as `decide (5 = 3) = false`.
2. Unsigned coercion of `a/b_offset_imm0`, where `rom.rs:226-236` reinterprets as `i64` and negates.
3. `a_imm1`/`b_imm1` ungated, where `rom.rs:240-244` zeroes them unless `a_src == SRC_IMM`.

## The correction this PR makes to its own first draft

The first version of this module claimed more than it proved, and adversarial review falsified it. Five of eleven slots (`line`, `a_imm1`, `b_imm1`, `ind_width`, `op`) were written **identically** in both forms, or shared the `romOpField` helper. On those the equality holds by `rfl`, so the cross-check had **no power at all** — the reviewer deleted the Fcall→CopyB remap from the shared helper and `romRowOf_eq_romRowSpec` still proved, while the docstring asserted that exactly such a slip "fails that proof instead of passing silently."

That is the failure mode the discipline exists to prevent, because the realistic error is one misreading of `rom.rs` transcribed into *both* forms — the second having been written by looking at the first.

Fixed: `romRowSpec` now writes every branch-carrying slot in a different shape — the `SRC_IMM` gates multiplicatively rather than as an `if`, the Fcall remap inline rather than through the shared helper, signed slots via `BitVec.toInt` rather than the `if v >= 0` split. Re-verified by mutation, each reverted:

- dropping the Fcall remap is now caught **by the agreement theorem itself** (unsolved goals at `romRowOf_eq_romRowSpec`), where before it was caught by nothing;
- dropping the `SRC_IMM` gate from both forms is caught.

Both docstrings now state exactly which slots are cross-checked, and record that `line` and `ind_width` are plain casts with no branch structure so nothing is claimed for them.

## What the agreement theorem does and does not say

**Says:** the `rom.rs` branch structure, the `zisk_inst.rs` bit chain, and the in-tree `packFlags` / `BitVec.toInt` readings coincide for every lowered row. `packFlags` is the PIL-mirroring packing every committed witness row is already written against — it was not written for this proof, so the check is against an independently maintained definition.

**Does not say:** anything about `trace.program`, the raw word, the ROM layout, or the extracted lowerer — none of which appear in the statement.

## Residuals

- **S3 (inhabitation on a concrete raw word) not attempted** — explicitly out of scope for this spike. Until it lands, this is a verified serialization, not yet a binding.
- `extract_transpile_rv64im_raw` hard-codes `rom_address := 0#u64`, so the extracted row's `paddr` is `0` while `romRowOf` reads it. S3 must confront this.
- Divergence 4 (the dropped Fcall→CopyB remap in June's version) is benign only if RV64IM provably never emits opcodes 246/247/248. Not proved here.
- The module is not in `extractionNamespaces` (`bin/TrustGate/Main.lean`), so V2 check 19/19 does not cover it even though it consumes the extraction. June's `RawProgramBinding` was added to that list for exactly this reason. Kernel-only closure was verified by hand instead.

**Kill criterion not hit** — no honest constraint-satisfying ROM row was found that is provably not the lowering of any 32-bit word. Route (c) survives this spike.

## Gates

- `lake build` — green, 9119 jobs, default target; module imported from `ZiskFv.lean`
- `check-all.sh` — 18/18 (689 modules reachable, 0 unreachable, 0 dangling)
- `check-all-semantic.sh` — 19/19

No new axioms/`sorry`/`native_decide`; all principal theorems close over `{propext, Classical.choice, Quot.sound}`.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -196,6 +196,11 @@ import ZiskFv.Compliance.AeneasBridgeTrust.Extraction
 -- (`zisk_decoder_accepts_supported_shape`) + completeness-obligation discharge
 -- (`real_decoder_accepts_in_shape`), proven in-build off the real Aeneas decoder.
 import ZiskFv.Compliance.AeneasBridgeTrust.Decode
+-- eth-act/zisk-fv#61 Phase-0 step S2: the cited line-by-line transcription of
+-- ZisK's ROM-row emitter (`rom.rs:204-260`) and its agreement with the in-tree
+-- PIL-mirroring `packFlags` / `BitVec.toInt` readings. Standalone; nothing in
+-- `root_soundness` depends on it yet (the raw-word binding is Phase-0 step S3).
+import ZiskFv.Compliance.RomRowSerialization
 import ZiskFv.Soundness
 import ZiskFv.Completeness.Rv
 -- Quarantined aspirational route; kept here so it stays built/verified, but the
--- a/ZiskFv/Compliance/RomRowSerialization.lean
+++ b/ZiskFv/Compliance/RomRowSerialization.lean
@@ -1,0 +1,536 @@
+/-
+ZiskFv/Compliance/RomRowSerialization.lean  (eth-act/zisk-fv#61, Phase-0 step S2)
+
+Line-by-line transcription of ZisK's ROM-row emitter,
+`zisk/state-machines/rom/src/rom.rs:204-260` (`RomSM::compute_trace_rom`), together
+with its `zisk/core/src/zisk_inst.rs:289-308` (`ZiskInst::get_flags`) flag packing.
+
+WHY THIS FILE EXISTS.  An accepted ZisK trace commits to a *lowered* eleven-slot
+`ZiskRomMessage` per instruction, never to a 32-bit RISC-V word.  Any future
+raw-word binding (`trace.program k = <serialization of the lowering of word k>`)
+is only as strong as the serialization it names, and a plausible-looking but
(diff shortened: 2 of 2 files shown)


codygunton:
Superseded by the phase9-* stack: the ROM-row emitter transcription was folded into `TraceLevelExport/RawProgramBinding.lean` (#313), now built directly on the production `packFlags` and the symbolic `SRC_IND`/`SRC_IMM` constants. Closing in favour of the stack.

Issue #304, opened 2026-07-29
Round-trip the hand-written mirrors against the generated constraints


codygunton:
Welding (#296, PR #300) ties a hand-written mirror clause to a generated constraint one theorem at a time. That catches transcription slips in whatever you weld, but it cannot find a constraint nobody welded — a missing weld fails nothing.

The 2026-07-28 fan-out demonstrated the gap concretely. Welding six AIR groups surfaced findings that only appeared *because* someone happened to work on that AIR:

- **Main `constraint_3` / `constraint_9`** — the a-side C-copy at `main.pil:385` — have no mirror counterpart anywhere. `sourceCCopyBetween` (`ZiskFv/AirsClean/Main/Circuit.lean:721`) models only the b-side.
- **MemAlign `cyclicSuccessorTransitionRows`** (`ZiskFv/AirsClean/MemAlign/Circuit.lean:255`) carries a ninth conjunct with no F-only generated counterpart — a mirror asserting *more* than the AIR.
- **MemAlign `constraint_16`** treats a column the AIR declares fixed (`mem_align.pil:120`) as a witness column.
- **`RomBoolSpec`** (`Main/Circuit.lean:346`) had no consumers anywhere in the tree — a published 14-conjunct mirror predicate checked against nothing.

None of these were found by a gate. They were found by hand, one AIR at a time.

## Approach: reuse #303's machinery in the other direction

#303 builds an AST and a polynomial canonicalizer for the pilout algebra. Apply the same thing to the mirror side: `g′ : assertZero body → Expression`, canonicalize, and compare the mirror's constraint set against the generated set per AIR.

Then coverage is derived, not declared:

- generated constraint with no canonical match in the mirror → **gap** (Main's a-side C-copy)
- mirror clause with no canonical match in the generated set → **strengthening**, which per `AGENTS.md` needs a source citation and a constructibility argument (MemAlign's ninth conjunct)
- lane-kind disagreement between mirror and the extractor header → **reclassification** (MemAlign `constraint_16`)

## Why this and welds are not redundant

A weld is an in-Lean `Iff`, so it is kernel-checked and composes with proofs. This check is external and only reports. They cover different failures:

- weld catches: the mirror I welded disagrees with the generated constraint
- this catches: a generated constraint no mirror models, or a mirror clause no generated constraint backs

The `Iff` direction matters for the second one — an implication-only weld (`spec_carryChain_weld` on the #300 branch) is blind to a mirror asserting more than the AIR.

[...]

PR #305, opened 2026-07-29
Weld the Main AIR mirror to the generated extraction

codygunton:
Part of #296. Welds the Main AIR mirror to the generated extraction.

**29 of 144 generated constraints welded**, and the split is the interesting part:

- `constraint_0..38` (39) are the real Main AIR local constraints (`main.pil:86..508`).
- `constraint_39..143` (105) are logup bookkeeping (`std_sum.pil`), reading `challenge` and stage-2 columns. Outside the F-only slice, same structural exclusion as Arith's 16.
- Of the 39 real ones, 29 are pure stage-1 row-local and are welded. The other 10 read `preprocessed`, `exposed`, or a rotation.

On the weldable slice the mirror is **exactly 1:1** with the generated file — same lanes, coefficients and association. Welds are `Iff.rfl`; predicate-level welds against `Spec`, `RomBoolSpec`, `SourceSpec` and `AddressSpec` need `simp only [sub_eq_add_neg]`-class normalization first, which was verified load-bearing (the raw `Iff.rfl` is rejected on `RomBoolSpec`'s `1 + -x` form).

`Extraction/Main.lean` had never been compiled — it was not in the build graph. It typechecks (0 errors, 667 KB olean) and is now built.

## Mirror findings (reported, not fixed)

- **Coverage gap:** generated `constraint_3`/`constraint_9`, the a-side C-copy at `main.pil:385`, have no mirror counterpart anywhere. `sourceCCopyBetween` (`Main/Circuit.lean:721`) models only the b-side. This is a real gap, independent of welding.
- **Structural, self-documented:** `constraint_4`/`constraint_10` (b-side C-copy) can never support an `Iff` — `sourceCCopyBetween`'s own docstring records that it deliberately drops the `__L1__`-selected public-input branch.
- **`constraint_18`** (pc handshake) substitutes the witness field `core.segment_l1` for the generated `preprocessed(0)`. The fixed-column binding does exist (`mainFixedColumns`/`mainFixedValues`), so this is not a defect, but it is not an `Iff.rfl` fact either.
- **Not a bug:** `addr0 - a_offset_imm0` and `addr2 - …` (`Main/Constraints.lean:268,270`) have no generated counterpart because those are not AIR witness columns; PIL inlines them into the mem-bus emission.
- **`RomBoolSpec`** had zero consumers in the tree before this module — a published 14-conjunct predicate checked against nothing. Now pinned.

## Hazard and lanes

The #300 binding hazard is closed here the same way: `extractedMainRowCircuit_pinned` via `inferInstance`, last in the module. `ReadsOnlyModeledLanes` holds by `Iff.rfl` for welded constraints and is **rejected** for `constraint_0` and `constraint_38`, so the stub lanes provably cannot be what makes a weld true.

Mutation-tested: mirror lane swap (`c_0`→`c_1`), operand swap (`a_0`→`a_1`), and two column-map mutations, each breaking the build with the exact `Iff.rfl` type mismatch recorded at the declaration.

**Known limit:** a column-map slip is caught only where that column is read by a welded constraint, and a compensating mirror+map slip still `rfl`s. That is what #304 exists to close; gate wiring belongs to the `weld-gate` track, not here.

Gates: `lake build` 9120 jobs, `check-all.sh` 18/18, `check-all-semantic.sh` 19/19. No new axioms/`sorry`/`native_decide`. No mirror edited.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -216,3 +216,9 @@ import ZiskFv.Regression.SignedDivOrdinaryCounterexample
 -- constraint definitions `@[simp]`, so importing it from anywhere with ZiskFv
 -- dependents would add them to the default simp set for every downstream module.
 import ZiskFv.AirsClean.ArithMirrorWeld
+-- `rfl`-level weld of the Main constraint mirror to the generated
+-- `Extraction.Main.constraint_N_every_row` predicates. This is the only ZiskFv
+-- module that may import `Extraction.Main`: that file marks all 144 generated
+-- constraint definitions `@[simp]`, so importing it from anywhere with ZiskFv
+-- dependents would add them to the default simp set for every downstream module.
+import ZiskFv.AirsClean.MainMirrorWeld
--- a/ZiskFv/AirsClean/MainMirrorWeld.lean
+++ b/ZiskFv/AirsClean/MainMirrorWeld.lean
@@ -1,0 +1,662 @@
+import Extraction.Main
+import ZiskFv.AirsClean.Main.Circuit
+
+/-!
+# Machine-checked weld: Main constraint mirror ↔ generated extraction
+
+`ZiskFv/AirsClean/Main/Constraints.lean` asserts a handwritten mirror of the
+Main AIR's row-local constraints, and `ZiskFv/AirsClean/Main/Spec.lean` /
+`ZiskFv/AirsClean/Main/Circuit.lean` publish the predicates the opcode proofs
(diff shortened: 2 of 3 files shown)


codygunton (commented):
Reviewed after rebase at f787b935. No blocking findings in the 29 row-local welds, live Main assertion projections, lane-restriction checks, or instance pin. Focused Lean build passed. The explicitly documented a-side C-copy and segment-boundary omissions remain scope decisions, not hidden claims in this PR.

PR #306, opened 2026-07-29
Weld the MemAlign AIR mirror to the generated extraction

codygunton:
Part of #296. Welds the MemAlign AIR mirror to the generated extraction.

**33 of 40 generated constraints welded.** All 33 are term-identical to the generated bodies — association, precedence, coefficient literals and conjunct ordering all match.

The #300 binding hazard is closed here the same way (instance pinned via `inferInstance`, last in module), and lane restriction is proved rather than assumed.

## Two findings, reported not fixed

**1. `constraint_16` reclassifies a fixed column as a witness column.** The AIR declares `col fixed L1 = [1,0,…]` at `mem_align.pil:120`; the mirror reads it as a witness column. Main does the analogous substitution for `SEGMENT_L1` and there it is justified by a real fixed-column binding (`mainFixedColumns`/`mainFixedValues`); whether MemAlign has an equivalent binding needs checking before this is called benign.

**2. `cyclicSuccessorTransitionRows` asserts more than the AIR.** `ZiskFv/AirsClean/MemAlign/Circuit.lean:255` carries a ninth conjunct, `current.delta_pc = successor.pc - current.pc`, with **no F-only generated counterpart**. Per `AGENTS.md` a mirror strengthening needs a source citation and a constructibility argument. This is the direction that can make real traces unprovable rather than unsound, so it wants a ruling.

Both are exactly the class #304 is meant to find mechanically instead of by hand.

MemAlign also carries two named defects (narrow-load lane, skippable-prove). The welds do not obscure them: no weld closes by assuming a defective row is impossible.

Gates: `lake build` 9120 jobs, `check-all.sh` 18/18, `check-all-semantic.sh` 19/19. No mirror edited; no new trust markers.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -222,3 +222,11 @@ import ZiskFv.AirsClean.ArithMirrorWeld
 -- constraint definitions `@[simp]`, so importing it from anywhere with ZiskFv
 -- dependents would add them to the default simp set for every downstream module.
 import ZiskFv.AirsClean.MainMirrorWeld
+-- `rfl`-level weld of the two MemAlign constraint mirrors (`Valid_MemAlign` and
+-- the Clean component's `Spec`/`transitionRows`/`cyclicSuccessorTransitionRows`)
+-- to the generated `Extraction.MemAlign.constraint_N_every_row` predicates. This
+-- is the only ZiskFv module that may import `Extraction.MemAlign`: that file marks
+-- all 40 generated constraint definitions `@[simp]`, so importing it from anywhere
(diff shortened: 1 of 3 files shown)


codygunton (commented):
Reviewed after rebase. No blocking implementation findings: both mirror bridges cover constraints 0–32, preserve row indexing, connect to the live component, and prove lane/window restrictions with pinned instances. Focused Lean build passed. The documented fixed-as-witness lane and extra ninth cyclic conjunct remain explicit model-scope decisions.

PR #307, opened 2026-07-29
Weld the MemAlign byte-family AIR mirrors to the generated extraction

codygunton:
Part of #296. Welds the MemAlignByte / MemAlignReadByte / MemAlignWriteByte AIRs.

**20 of 41 generated constraints welded.** One shared bridge across the three AIRs rather than three near-copies.

The #300 binding hazard is closed the same way; lane restriction proved, not assumed. Mutation-tested with the exact errors recorded at the declarations they break.

The unwelded remainder is stated per constraint in the module — the excluded ones read lanes outside the F-only slice or have no 1:1 mirror clause.

Gates: `lake build` 9122 jobs, `check-all.sh` 18/18, `check-all-semantic.sh` 19/19. No mirror edited; no new trust markers.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -230,3 +230,10 @@ import ZiskFv.AirsClean.MainMirrorWeld
 -- with ZiskFv dependents would add them to the default simp set for every
 -- downstream module.
 import ZiskFv.AirsClean.MemAlignMirrorWeld
+-- `rfl`-level weld of the three `mem_align_byte.pil` constraint mirrors to the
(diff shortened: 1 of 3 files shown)


codygunton (commented):
Reviewed after rebase. No blocking findings: all 20 F-only constraints across the byte family weld at their AIR-specific layouts, the two live Clean assertion lists are connected, lane restrictions are explicit, and all three instances are pinned. Focused Lean build passed. The WriteByte surface is accurately described as template evidence without a live consumer.

PR #308, opened 2026-07-29
Weld the Binary-family AIR mirrors to the generated extraction

codygunton:
Part of #296. Welds the Binary / BinaryAdd / BinaryExtension AIR mirrors.

**11 of 31 generated constraints welded.** The survey found the coverage asymmetry expected from the mirror densities: BinaryExtension has ~3 `assertZero` against 8 generated constraints, Binary has ~19 against 14. Both directions are reported per constraint in the module rather than smoothed over.

The #300 binding hazard is closed the same way; lane restriction proved; mutations recorded at the declarations they break.

Gates: `lake build` 9121 jobs, `check-all.sh` 18/18, `check-all-semantic.sh` 19/19. No mirror edited; no new trust markers.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -133,6 +133,11 @@ import ZiskFv.EquivCore.WriteValueProofs.MulDivRemSigned
 import ZiskFv.EquivCore.Bridge.MemClean
 import ZiskFv.AirsClean.ArithTableProjections
 import ZiskFv.AirsClean.BinaryFamily.Balance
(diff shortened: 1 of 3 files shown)


codygunton (commented):
Reviewed after rebase. No blocking finding in the inverse `validOfCircuit` weld design: it avoids instance/stub hazards, covers all 11 F-only Binary/BinaryAdd constraints, and accurately limits its claim for challenge-mixing columns. Focused Lean build passed. Follow-up gate review found that #310 must explicitly discover and validate this inverse-map form; that fix will land in #310 rather than weakening this design.

PR #309, opened 2026-07-29
Weld the Mem AIR Spec mirror to the generated extraction

codygunton:
Part of #296. Welds the Mem AIR mirror to the generated extraction.

**Read the increment carefully — the first version of this branch overstated it, and the correction is committed in `6bba1378`.**

## What is genuinely new

`Extraction.MemGeneratedConstraintBridge` already exists, is auto-generated, binds `Extraction.Mem.constraint_0..33`, and is a **`defaultTargets` entry in `lakefile.toml`** — so it has been gating the Mem mirrors all along. The `segment_weld` and `permutation_weld` families here largely duplicate it.

The genuine increment is:

- **Part A** — `spec_weld` and `extracted_of_componentWithDualMemBus_constraints`, over the `Spec`/`main` mirror the generated bridge does not cover (9 constraints).
- `∀`-generality over the single `proverData` instantiation the bridge is stated at.
- The lane-restriction result and instance pinning.

## Why the original claim was wrong, since it generalizes

The module originally asserted that a self-consistent slip across two hand-written mirrors is caught by `permutation_weld` "and nothing else in the build". Adversarial review reproduced the mutation and found the generated bridge rejects it too, at `:266` and `:515`, with this module absent from that target's closure.

Root cause: the mutation runs used the targeted `lake build ZiskFv.AirsClean.MemMirrorWeld` (8203 jobs), which excludes the bridge (job 8986 of 9119 in the default build). *"Not in my target's import closure"* was written up as *"provably cannot see it"*. **Any mutation conclusion drawn from a targeted build inherits that error** — measure against the default build.

A second false claim, also corrected: the module cited `check-arith-column-map.py` as existing precedent; it exists only on the unmerged `arith-mirror-weld` branch.

## Note on review history

A verifier caught this branch's worktree dirty with an unreverted mirror mutation (`constraint 33`'s `l1 (row + 1)` → `l1 row`) that the agent had reported as reverted. Both mirror files are now confirmed **byte-identical to `origin/main`**, and the branch adds only new files. Worth knowing when reviewing.

Gates: `lake build` 9119 jobs, `check-all.sh` 18/18, `check-all-semantic.sh` 19/19. No mirror edited; no new trust markers.

Open hole, disclosed: the four lane maps are handwritten from the extractor header with nothing pinning them, so a compensating mirror+map slip still `rfl`s. That is #304.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -242,3 +242,9 @@ import ZiskFv.AirsClean.MemAlignMirrorWeld
 -- importing one from anywhere with ZiskFv dependents would add them to the
 -- default simp set for every downstream module.
 import ZiskFv.AirsClean.MemAlignByteMirrorWeld
+-- Machine-checked weld of the three Mem constraint mirrors against the generated
+-- `Extraction.Mem` constraint set. Deliberately a leaf, imported last: it is the
+-- only module under `ZiskFv/` that imports `Extraction.Mem`, whose 34 generated
+-- constraint definitions are all `@[simp]`-tagged, so keeping it out of every
+-- other module's import closure keeps that simp set out of the proofs.
+import ZiskFv.AirsClean.MemMirrorWeld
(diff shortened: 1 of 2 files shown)


codygunton (commented):
Reviewed after rebase at the corrected increment. No blocking findings: all 34 generated constraints are welded, the live nine-constraint Mem component path is connected, row-sensitive lanes and instance bindings are checked, and the overlap with the pre-existing generated bridge is stated honestly. Focused Lean build passed.

PR #310, opened 2026-07-29
Make the weld column-map gate AIR-parameterised

codygunton:
Stacked on #300 (base `arith-mirror-weld`). Part of #296.

Makes the weld column-map gate **AIR-parameterised**, so registering a new welded AIR is a data change, not a script change.

Five sibling weld branches (#305 Main, #306 MemAlign, #307 MemAlign-byte, #308 Binary, #309 Mem) were deliberately forbidden from touching `trust/scripts/`, precisely so they would not each add a check numbered `N/19` and turn the landings into a conflict queue. This is the track that owns that wiring.

- One check in `check-all.sh` regardless of how many AIRs are welded.
- Per-AIR column maps recorded as data under `trust/generated/weld-columns/` (`arith.txt` today).
- **Fail-closed preserved.** `.github/workflows/trust-gate.yml` checks out *without* `nix run .#populate`, so a SKIP-on-missing-input branch would mean the gate checks nothing on every push to main. Inputs are asserted; a missing or empty recording exits non-zero with `refusing to pass vacuously`. When the generated file is present, the recording must reproduce it exactly.
- Data format documented in the script header so the sibling branches can conform as they land.

Mutation-tested: corrupting a recording fails; removing a recording fails.

Gates: `lake build`, `check-all.sh`, `check-all-semantic.sh` — counts in the commit.

Note for sequencing: this and the five weld PRs all touch `lakefile.toml` and `ZiskFv.lean`, so landings serialize and each needs a re-gate after rebase.


codygunton (commented):
Re-reviewed after the requested completion at 69e52b92. The earlier blocker is fixed: discovery now covers both every `*MirrorWeld.lean` file and every circuit instance; Binary/BinaryAdd use a checked inverse-map schema; all forward carriers are independently registered and tied through their kernel pin/helper; 11 carrier maps reproduce nine generated AIR headers, including explicit Mem aliases. Final gates: `lake build` 9,132 jobs PASS; V1 19/19 PASS; V2 semantic 19/19 PASS. Approved for merge.

Issue #303, opened 2026-07-29
Round-trip the extractor: check every pilout constraint reaches Lean unaltered


codygunton:
Today the extractor `tools/pil-extract` is trusted end to end: nothing checks that every constraint in `build/zisk.pilout` reaches `build/extraction/Extraction/<AIR>.lean`, or that it arrives unaltered.

Counting emitted constraints, or diffing the `-- <air>.pil:NNN` provenance comments, only checks *accounting*. Both pass while a constraint is silently mistranslated — right count, right cited line, wrong polynomial.

## Approach: round-trip in the source algebra

Both sides are already structured. `tools/pil-extract/proto/pilout.proto` defines `Constraint` and `Expression` messages; the emitted Lean is machine-regular (`Extraction.Circuit.main c (id := 1) (column := 9) (row := row) (rotation := 0)` plus field arithmetic).

With `f : pilout → Lean` the extractor, build `g : Lean → Expression` and check `g(f(t)) ≡ t` for every `t`, entirely in the pilout's algebra.

- A dropped constraint shows up as `g(f(T))` failing to cover `T`.
- A distorted one shows up as a non-equivalence.

Coverage falls out of the data rather than out of a maintained declaration list.

## `≡` should be polynomial normal form, not a rewrite list

Both sides are polynomial expressions over column atoms, so canonicalize each to a sorted monomial list with Goldilocks coefficients and compare exactly. That is total and decides the question; an enumerated set of allowed rewrites (association, commutativity, `a - b` vs `a + (-1)*b`, constant folding) is strictly weaker and needs maintaining.

Random evaluation has a role, but as a **fast screen** rather than the decider: evaluate both forms at random assignments to locate candidate mismatches cheaply, then canonicalize only the suspects. This matters if Arith's carry chains make full expansion expensive.

## Steps

- [ ] **P0.** Parse the pilout; count constraints per AIR; diff against the emitted `constraint_N_every_row` set. Read-only, no build. Establishes whether anything is dropped at the count level.
- [ ] **P1.** `g`: parse emitted Lean bodies into the proto `Expression` AST.
- [ ] **P2.** Canonicalizer + comparator. Report per AIR: matched / mismatched / pilout-only (dropped) / lean-only (invented).
- [ ] **P3.** Gate it after extraction in `nix run .#populate` and CI. Fail on any mismatch or uncovered pilout constraint.

## Two conditions on the construction

[...]

PR #20, opened 2026-05-10
Step 2b prep: Binary AIR range-check axiom

codygunton:
## Summary

Adds `binary_columns_in_range` axiom packaging the per-row range bounds for `Valid_Binary`'s `bits(N)`-annotated columns. PIL: `binary.pil:60-86`. Trust ledger 88 → 89.

Trust class: lookup-argument soundness on the standard range-checker bus, mirroring `binary_add_columns_in_range` (#19 / Step 1.5), `main_columns_in_range` (Step 1 / now in #19), and `bin_table_consumer_wf` (BinaryTable bus). Same scope as user's accepted SNARK-protocol category.

## Scope honesty: infrastructure-only, not *promise discharge*

Per the *anti-laundering principle*, this PR is explicitly the prerequisite for the next PR's *promise discharge* work. The *anti-laundering metric* HOLDS (no canonical theorems touched) — which the gate allows but the principle flags as not-yet-real-progress.

Stages to complete Step 2b after this PR:

1. Author the Binary AIR *discharge bridge* (`ZiskFv/Equivalence/Bridge/Binary.lean`) consuming `op_bus_perm_sound_Binary` + `binary_columns_in_range` + `bin_table_consumer_wf` + `Valid_Binary` structure.
2. Refactor 1+ of the 14 Binary-shape opcodes (AND/ANDI/OR/ORI/XOR/XORI/SUB/SUBW/SLT/SLTI/SLTU/SLTIU/ADDIW/ADDW) to consume the bridge. Each refactor should reduce per-theorem hypothesis counts measurably.

The bridge + first-opcode discharge fits one focused follow-up PR. Cleanly stacks on this one.

## Why split (vs. bundling with the bridge)

The bridge work is ~3-5 hours of build/iterate given Lean cycle costs. Authoring + landing the axiom separately keeps the bridge PR focused on the actual *promise discharge* work and its anti-laundering metric reduction, rather than mixing axiom plumbing with proof refactoring.

This PR stacks directly on #19. Build off it for the bridge.

## Test plan

- [x] `lake build ZiskFv.Airs.Binary.BinaryRanges` passes.
- [x] V1 trust gate: all 8 checks pass.
- [x] Baselines unchanged (hypothesis-count, caller-burden); no canonical theorem refactored.
- [ ] V2 (`check-all-semantic.sh`) deferred to CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #19, opened 2026-05-10
Step 1.5: real *promise discharge* on equiv_ADD via BinaryAdd bridge

codygunton:
## Summary

Concretely demonstrates the *anti-laundering metric* (V1 checks #7 + #8) shrinks under a real *promise discharge*. Replaces Phase A's pilot, which laundered ADD's caller-burden (45 → 47 binders) by relabeling, with a derivation rooted in the *trust ledger*.

ADD's caller-burden journey:

| State | total | hypothesis | Δ vs origin/main |
|--|--:|--:|--|
| origin/main pre-pilot | 45 | 30 | — |
| Phase A pilot (PR #16) | 47 | 32 | +2 / +2 ⚠ LAUNDERED |
| **this PR** | **43** | **29** | **−2 / −1 ✓ DISCHARGED** |

Aggregate (across all 63 canonical theorems): 3474 → 3472 total, 2322 → 2321 hypothesis.

## What's in this PR (vocabulary per [`docs/fv/known-gaps.md` Glossary](https://github.com/eth-act/zisk-fv/blob/main/docs/fv/known-gaps.md#glossary-canonical-terminology))

- `ZiskFv/Airs/Binary/BinaryAddRanges.lean` (NEW) — one new axiom `binary_add_columns_in_range`, the BinaryAdd analogue of Step 1's `main_columns_in_range`. Trust class: range-check bus lookup soundness on `bits(N)`-annotated columns. PIL: `binary_add.pil:7-10`.
- `ZiskFv/Equivalence/Bridge/BinaryAdd.lean` (NEW) — the *discharge bridge*. Consumes Phase A's `op_bus_perm_sound_BinaryAdd` (PLONK soundness) + Step 1.5's `binary_add_columns_in_range` + `Valid_BinaryAdd` to produce the existential row witness + `add_circuit_holds` + range facts + per-byte input bridges that `RdValDerivation.Arith.h_rd_val_arith_add` consumes.
- `ZiskFv/Equivalence/Add.lean` — `equiv_ADD` refactored to consume the bridge. Drops `r_binary`, `h_circuit`, `h_a_range/b/c`, `h_input_r{1,2}_circuit`. Adds `h_main_subset`, `h_main_mode`, universal `h_b_core`, Main-form `h_input_r{1,2}_main`. Net: 7 binders REMOVED, 5 ADDED, −2 binders.

This PR also folds in source files from prior unmerged work it depends on:
[...]

PR #319, opened 2026-07-30
Pair every generated constraint against the mirrors that restate it

### LINKED ISSUE #304: Round-trip the hand-written mirrors against the generated constraints

codygunton:
Closes #304. **Stacked on #312** (base is `s11-roundtrip`); merge that first.

Welding ties one mirror clause to one generated constraint and is kernel-checked, but a missing weld
fails nothing. Every mirror gap found on 2026-07-28 surfaced because a person happened to work that
AIR. This applies #303's canonicaliser in the other direction: pair each comparable generated
constraint against the mirror clauses under `ZiskFv/AirsClean/**` that restate it, and classify every
difference.

## ⚠️ This step FAILS at HEAD, and that is the deliverable

```
mirror-roundtrip: FAILED 139/176 matched; 35 gap, 2 strengthening, 3 unbacked,
                  2 reclassification, 1 unreachable mirror, 0 unparsed
acceptance:       OK, 16 cases -- 4/4 hand findings reproduced, 8/8 mutations
                  classified as predicted, 3/3 neutral controls, 1 measured blind spot
```

**Merging this reds `nix run .#test` until the findings are closed or declared.** That is a decision
for you, not for me, and it is the one thing in this PR I would not merge without a ruling. The two
defensible options:

- **(a) as written.** The gate is a hard failure at step 4/10. Maximum pressure, red main.
- **(b) a citation-bearing baseline that can only shrink**, in the idiom this repo already uses for
  `check-shrinkage.sh` and `check-defect-count.sh`. The 42 current findings become a checked-in
  declared list, the gate goes green, and any *new* gap fails. The issue's own text authorises this
  ("either fixed or explicitly declared with a citation, and the declaration list must be small
  enough to read").

I did not build (b), because the instruction to the implementer was not to silence findings and I did
not want that judgement made by an agent. Say the word and it is a small follow-up.

## Per-AIR

| air | comparable | mirror | matched | gap | strengthen | recl | unbacked |
|---|---:|---:|---:|---:|---:|---:|---:|
| Arith | 49 | 60 | 49 | 0 | 0 | 0 | 0 |
| Binary | 7 | 14 | 7 | 0 | 0 | 0 | 0 |
| BinaryAdd | 4 | 8 | 4 | 0 | 0 | 0 | 0 |
| BinaryExtension | 0 | 0 | 0 | 0 | 0 | 0 | 0 |
| **Main** | 39 | 42 | 29 | **9** | 2 | 1 | 2 |
| **Mem** | 24 | 18 | 9 | **15** | 0 | 0 | 0 |
| MemAlign | 33 | 33 | 32 | 0 | 0 | 1 | 1 |
| **MemAlignByte** | 9 | 10 | 5 | **4** | 0 | 0 | 0 |
| MemAlignReadByte | 4 | 5 | 4 | 0 | 0 | 0 | 0 |
| **MemAlignWriteByte** | 7 | 0 | 0 | **7** | 0 | 0 | 0 |
| **total** | **176** | **190** | **139** | **35** | 2 | 2 | 3 |

**The 35 gaps are not all the same thing, and the headline count overstates.** 15 of them are Mem's,
and each is annotated `out-of-root claim: segmentResidualEveryRow ZiskFv/Airs/Mem.lean:296 claims it`
— that predicate exists, is documented as "the generated segment constraints not already enforced by
the nine local Mem component equations", and `:334-339` destructures a 24-conjunct
`segment_every_row` into exactly those 15. So Mem is modelled, just split across `Airs/` and
`AirsClean/`, and the declared mirror root (`ZiskFv/AirsClean`, from the issue's own wording) does not
see it. Either widen the root or declare the split. **Constraints nothing models anywhere: 20.**

## The findings

**Main #3 and #9 — the flagship, confirmed independently.** `main.pil:385`
`(a_src_c)*(a[i]-(previous_c))`. `grep a_src_c ZiskFv/` returns **zero files**: unmodelled tree-wide,
not merely absent from the mirror. `sourceCCopyBetween` covers only the b-side #4/#10 at `:386`.

**A second unmodelled Main cluster the hand fan-out did not name** — the segment boundary: #19
`SEGMENT_L1'*(segment_next_pc-(next_pc'))` (`:423`), #20/#21 `SEGMENT_L1'*(segment_last_c[i]-c[i])`
(`:426`), #38 `SEGMENT_L1*(segment_initial_pc-pc)` (`:508`), and #0
`main_last_segment*(1-main_last_segment)` (`:86`).

**MemAlignWriteByte has no mirror at all** — zero mentions in `ZiskFv/` or `trust/`, and no
`AirsClean/MemAlignWriteByte/` beside `MemAlign/`, `MemAlignByte/`, `MemAlignReadByte/`. Easy to
misread, so stating it precisely: all three byte-family AIRs instantiate one PIL template
(`mem_align_byte.pil:35,36,37,59,83,87,88`), so terms like `written_composed_value` *do* appear in the
tree — in `AirsClean/MemAlignByte/`, over that AIR's columns. Stage widths differ per AIR (16+4 /
10+3 / 14+4), so those are different polynomials and genuinely do not cover the write-byte AIR. Given
the family provides tuples for sub-doubleword *loads*, a loads-only scope may well be deliberate —
but nothing declares it, which is exactly what this issue turns into a declared exclusion.

**`RomBoolSpec` is unreachable, and 14 Main constraints are hollow.** Confirmed 0 external
references. The tool reports these separately rather than counting them as coverage: a clause of an
unreachable predicate can be canonically perfect and still constrain nothing, because no proof
consumes it.

**MemAlign's ninth conjunct (the issue's strengthening case) resolved mechanically**:
`cyclicSuccessorTransitionRows` has 9 conjuncts for 8 comparable constraints, and the extra
`delta_pc` clause's only generated source sits inside the *excluded* challenge-mixed
`constraint_36`. So it is a strengthening relative to the F-only slice specifically.

**`MemAlign.L1` is a fixed column modelled as the free field `preL1`** — the reclassification case —
and `MemAlign.component` declares no `fixedColumns` at all, unlike Main and Mem.

## Evidence the gate can fail, and one measured blind spot

`acceptance.py`: the 4 hand findings must be rediscovered without being told where to look, plus 8
mutations of a *copy* of the mirrors (delete a matching clause, add a plausible extra, repoint a
projection, swap a fixed reference for a witness, change a row variable, …), each required to move the
report in exactly the predicted way, plus 3 neutral controls (reorder conjuncts, restate `a = b` as
`a - b = 0`) that must not move it. The real tree's sha256 is checked unchanged at the end.

One blind spot is *measured* rather than claimed: deleting a mirror clause that a **second** mirror
also restates keeps the match, because pairing is set-to-set and there is no per-mirror coverage
count. That case passes, and the pass is recorded as a blind spot, not as coverage.

## Lane map with kind, which is what makes `constraint_16` mechanical

`lanes.py` derives, per AIR, name → lane from `PilOut.symbols` and the emitted header only —
never from a mirror and never from an existing handwritten map, since those are what is being gated.
Lanes carry KIND (stage-1 witness / stage-2 / fixed / exposed), and `AirValue` is kept apart from
`AirGroupValue` rather than inheriting #303's collapse. It is total and closed in both directions and
**raises** instead of defaulting.

That matters because the existing maps do not. `h998ExprToField`
(`ZiskFv/AirsClean/MemAlign/Bridge.lean:31`, 38 arms) and `c10LookupExprToClean`
(`Binary/Wiring.lean:33`, 18 arms) both end in `| _ => 0`, each gated by a single `rfl` over one
tuple: a wrong or missing index becomes a silent zero, which weakens rather than fails. The survey
inventoried all 30 such maps in the tree; `Mem/RangeWiring.lean:29` is the one that refuses
(`| _ => none`) and is the pattern to copy.

## Declared exclusions — three categories, each cited, each listing what it took

`challenge_or_stage2` (179 generated; the issue's own comparable rule, implemented twice
independently and required to agree every run), `mirror_carrier_not_a_row` (5; honest-row builder
inputs and a component-owned fixed-column schema, which are inputs to a row rather than slots of one),
`mirror_clause_not_an_equation` (29; `.val` bounds and delegations whose delegate is compared
elsewhere — and a delegate that is *not* is reported as an undeclared delegation, of which there is
currently 1).

## Scope and trust

Polynomial identities only, same as #303. This tool **reports**; nothing under `ZiskFv/` or `trust/`
was modified (verified in the diff and by checksum in `acceptance.py`). The gaps and strengthenings
are real proof work and are not the tool's to fix — mirrors are protected proof interfaces.
No axiom, no ledger change, no TCB change. #303's gate still passes on this branch.

## Not verified

`nix run .#test` was not run end to end. `check_mirrors.py` (exit 1, as designed) and
`acceptance.py` (exit 0) were both run directly, as was `tools/pilout-roundtrip/check.py` (exit 0).

## Overlap to resolve before both land

PR #310 makes the weld column-map gate AIR-parameterised; `lanes.py` is the same gate with lane kind
added. The issue says #304 absorbs the column-map gate. Pick one rather than merging both.

### DIFF
--- a/.github/workflows/proofs.yml
+++ b/.github/workflows/proofs.yml
@@ -256,7 +256,7 @@ jobs:
           # proof-memory fix, and matches nix/test.nix's default.
           LEAN_NUM_THREADS: "4"
           # The independent Aeneas production-extraction completeness check
-          # (step 4/9) is the single largest cost (~1.5h) and is unrelated to
+          # (step 5/10) is the single largest cost (~1.5h) and is unrelated to
           # the Lake proof graph. Run it in the parallel
           # aeneas-production-extraction job below instead of serializing it
           # ahead of `lake build`. Plain local `nix run .#test` is unaffected.
@@ -322,7 +322,7 @@ jobs:
 
   # ------------------------------------------------------------------
   # Aeneas RV-completeness extraction harness, run in PARALLEL with the
-  # Lake FV gate above. Previously this ran as step 4/9 of `nix run .#test`
+  # Lake FV gate above. Previously this ran as step 5/10 of `nix run .#test`
   # inside the lake-build job, serializing ~1.5h of generated-extraction
   # work ahead of the proof build. It is independent of the Lake proof
   # graph (the `aeneas-extraction-diff` job already guards the tracked
--- a/docs/extraction/extractor-notes.md
+++ b/docs/extraction/extractor-notes.md
@@ -456,13 +456,76 @@ directory: an AIR whose constraints all became `--skip-unsupported` stubs has no
 perfect ratio.
 
 Wiring: the tail of `nix run .#populate` (so drift fails where it is produced)
-and step 3/9 of `nix run .#test` (`check.py` plus the mutation selftest). Not in
+and step 3/10 of `nix run .#test` (`check.py` plus the mutation selftest). Not in
 `trust/scripts/check-all.sh`, whose CI job has no `build/` at all. Exit 1 is a
 mismatch or an uncovered constraint, exit 2 is a missing artifact; neither is a
 pass.
 
 `tools/pilout-roundtrip/README.md` carries the full argument, the two atom
 vocabularies, and the measured residual blind spots.
+
+## Mirror gate (`tools/mirror-roundtrip`, eth-act/zisk-fv#304)
+
+The round-trip gate above ends at `build/extraction/`. Nothing in its argument
+touches `ZiskFv/`, and no Lean under `ZiskFv/` imports a per-AIR
+`Extraction.<AIR>` module at all, so a constraint can round-trip perfectly and
+still be restated wrongly, partially, or not at all in the handwritten Lean the
+proof actually consumes. This gate closes that second direction for the
+polynomial content.
+
+How the two differ:
+
+| | round-trip gate (#303) | mirror gate (#304) |
+| - | - | - |
+| decides | `pilout` vs `build/extraction/` | `build/extraction/` vs `ZiskFv/AirsClean/**` |
+| both sides are | generated | one generated, one handwritten |
+| a failure means | the extractor dropped or distorted a constraint | a constraint has no mirror, or a mirror has no constraint |
+| its verdict is | `OK` at HEAD | `FAILED` at HEAD, and the failures are the finding |
+| the fix is | change the generator and rerun | proof work on a protected interface |
+
+It reuses `poly.py`, `check.to_poly`, `lean_parse.py`, `pilout_wire.py` and
+`pilout_atoms.py` from `tools/pilout-roundtrip` rather than forking them, so the
+two tools cannot disagree about what a column, an atom or a canonical form is.
+Per AIR it canonicalises the *comparable* generated constraints — the issue's
+rule: all of them minus every constraint reaching `Extraction.Circuit.challenge`
+or a stage-2 lane, 176 of the 355 — and the clauses of every inventoried mirror
+predicate, and pairs the two sets by canonical form rather than by index. The
+comparable rule is implemented twice, off the emitted Lean and off the pilout
+operands, and the two index sets must agree on every run.
+
+Five finding classes: `MATCHED`; `GAP`, a comparable constraint no mirror clause
+has the canonical form of; `STRENGTHENING`, a mirror clause no constraint has the
+form of — a *syntactic* class, so the report also runs a cofactor search and says
+when the clause is in fact implied by a constraint and therefore weaker rather
+than stronger; `RECLASSIFICATION`, a pairing that turns on a lane's kind;
+`UNBACKED`, an equation over a row field this AIR has no lane for, which has no
+canonical form to pair with and so is reported rather than compared.
+
+Besides the pairing, the run holds its own scope: `survey.CLASSIFICATION` against
+the declarations actually under the mirror root, every `NEAR_*`-classified
+declaration re-parsed and required to carry no comparable equation,
+`DECLARED_AIRS` through #303's own `_check_scope`, `lanes.gate_lane_map`, every
+resolved projection against the row record it claims to project, and a non-empty
+floor on both denominators. A scope that shrinks is the failure mode a
+declared-list scope has, and each of those is what makes one loud.
+
+Wiring: step 4/10 of `nix run .#test` (`check_mirrors.py` plus `acceptance.py`,
+the mutation suite that is the evidence the gate can fail and classify). Exit 1
+is a finding, exit 2 a missing artifact; neither is a pass. Deliberately NOT in
+`nix run .#populate`: populate materialises generated inputs, and this gate's
+failing side is handwritten source that populate neither writes nor reads, so a
+gap here is not extraction drift and re-running populate cannot change it.
+Deliberately not in `trust/scripts/check-all.sh` either, for #303's reason — that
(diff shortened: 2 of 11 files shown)


codygunton:

PR #321, opened 2026-07-31
Mirror round-trip: see the out-of-root Mem mirror and Bool-typed columns

codygunton:
Follow-up to #319 (#304), **stacked on it** (base `s11-mirror-roundtrip`). Merge #319 first.

#319 reported `139/176` with 35 gaps. 19 of those 35 were false positives from how the *tool* is written — the mirrors are correct, the tool couldn't see them. This closes exactly those 19, honestly, by teaching the tool to see the coverage rather than declaring it away. Diff is confined to `tools/mirror-roundtrip/`; nothing under `ZiskFv/`/`trust/`/`nix/` changed (verified by checksum in `acceptance.py`).

## Result

`158/176 covered (139 matched, 15 out-of-root, 4 bool-typed); 16 gap` — down from 35.

| AIR | before | after |
|---|---|---|
| Mem | 15 gap | **0 gap, 15 OUT_OF_ROOT** |
| MemAlignByte | 4 gap | **0 gap, 4 BOOL_TYPED** |
| Main | 9 gap | 9 gap (unchanged) |
| MemAlignWriteByte | 7 gap | 7 gap (unchanged) |

The remaining 16 are all genuine and pre-existing: Main 9 (#3/#9 unmirrored a-side; #4/#10 the weakened b-side — see the findings doc in the sibling PR; #0/#19/#20/#21/#38 the #103/#76 segment boundary) + MemAlignWriteByte 7 (no mirror at all).

## Group A — 15 Mem: matched, just out of the searched root

`segmentResidualEveryRow` (`ZiskFv/Airs/Mem.lean:296`) states all 15; the tool's root was `ZiskFv/AirsClean` so it never parsed them. Now the tool parses that mirror, resolves its `SegmentColumns` fields through the **same `lanes.LaneMap`** (no second map), and pairs by canonical form. **All 15 are exact canonical matches — empty symmetric difference, no clause left over.** Reported as `OUT_OF_ROOT`, which is the same polynomial decision as `MATCHED`, only the mirror lives outside `AirsClean`. Not a declared exclusion — a real match.

## Group B — 4 MemAlignByte: discharged by typing, not by an equation

`sel_high_4b/2b/b` and `is_write` booleanity is discharged by the row field being `Bool`-typed / carrying a `.val < 2` bound (`MemAlignByte/Spec.lean:71-72`), so no polynomial clause exists to pair. The recognizer is **mechanical and evidence-bearing**, not an index list: a generated constraint is `BOOL_TYPED`-covered iff its canonical form is exactly `a*(1-a)` for a single stage-1 witness atom `a` **and** that column carries a Bool declaration or a `.val < N (N≤2)` bound whose subject resolves to `a`. It discriminates — **MemAlignWriteByte's identical selector booleans have no row record and no bound, so they stay gaps** (confirmed: still 7).

## Evidence

`acceptance.py` OK, 18 cases (was 16). Two added: `OUT_OF_ROOT_DISAGREES` (mutate a residual clause to disagree → the Mem constraint reverts to GAP) and `BOOL_TYPING_WEAKENED` (loosen `sel_high_4b`'s bound to `< 256` → MemAlignByte #0 reverts to GAP). Both new recognizers are thereby shown to fail when they should. Real `ZiskFv/`/`trust/` byte-identical before and after (`sha256 5f37e80c…`). `tools/pilout-roundtrip/check.py` still `355/355`.

Note for review: two pre-existing acceptance cases (`CLAUSE_DELETED`/`PROJECTION_SWAPPED`) were retargeted from selector booleans onto algebraic MemAlign clauses (#30 `sel_prove`, #31 `value_0`), because deleting a *bounded* selector's Spec clause now correctly leaves it BOOL_TYPED-covered. The agent states no case was removed or weakened; worth an eyeball since retargeting an acceptance case is the kind of move that can mask a regression.

## Still failing, by design

The gate still exits 1: 16 real gaps + 2 strengthening + 2 reclassification + 1 unreachable remain. Those are the owner's to close or declare — unchanged from #319. This PR only removes the 19 that were never real.


codygunton:

Issue #76, opened 2026-06-12
Reduce load memory premise to RowTraceCoherence floor


codygunton:
## Current status (2026-06-24)

Closed as resolved to the load-side trace-coherence floor.

The old whole-state load memory premise has been reduced to `LoadMemoryTimelineCoherenceEvidence`, whose core residual is `RowTraceCoherence` in `ZiskFv/ZiskCircuit/MemTimeline/Spike.lean`. The load rowData field is still named `h_memory_timeline`, but its type is now memory-map-only rather than a whole-`SailState` replay pin.

Current code pointers:

- `ZiskFv/Compliance/OpEnvelope.lean`: `LoadMemoryTimelineCoherenceEvidence` and `loadMemoryTimelineEvidence_of_coherenceEvidence`.
- `ZiskFv/Compliance/TraceLevelExport/RowDataArithMem.lean`: load `h_memory_timeline` fields now carry the reduced evidence.
- `ZiskFv/ZiskCircuit/MemTimeline/Spike.lean`: `RowTraceCoherence`, byte-local agreement derivation, and non-degeneracy witness.

What remains is tracked separately:

- #115: eliminate or justify the remaining `RowTraceCoherence` floor.
- #119: reduce the store RMW byte residuals `h_m1..h_m7` by an analogous store-arm refactor.

codygunton:
Superseded status note (2026-06-24): this intermediate P3 record predates PR #112/#113 and the terminology refresh. The current closed state is recorded in #76: loads use `LoadMemoryTimelineCoherenceEvidence` / `RowTraceCoherence`; floor elimination is #115 and store RMW residuals are #119.

codygunton:
Superseded spike note (2026-06-24): this blocked-state comment is historical. #103 is now closed as a banked cross-segment Mem seam capability; the current memory follow-ons are #115 for the trace-coherence floor and #119 for store RMW byte residuals.

codygunton:
Superseded #103 fold-in note (2026-06-24): this referenced old branch artifacts from the closed cross-segment Mem seam spike. The current #76 body records the load-side reduction; current follow-ons are #115 and #119.

codygunton:
[...]

PR #322, opened 2026-07-31
Investigate Main constraint_4/10: the weakened source-C copy

codygunton:
Follow-up to #319 (#304), **stacked on it** (base `s11-mirror-roundtrip`). Merge #319 first. Docs-only — one new findings file, no code, no proof, no tool change.

#319 flagged Main `constraint_4`/`constraint_10` as the subtle case: mirrored, but by a *weakened* copy. This is the investigation. Every claim below is verified from source (I re-checked the three load-bearing ones independently).

## Verdict: real, precisely characterised, soundness-safe. It is #103/#76 cross-segment scope — not an untracked hole, not unsound.

**1. The only difference is the segment factor (proved).** `main.pil:386` emits `b_src_c*(b[i]-previous_c) = 0` on every row. The mirror `sourceCCopyBetween` (`Main/Circuit.lean:721`) states `(1-segment_l1)*b_src_c*(b[i]-prev.c) = 0`. Since `SEGMENT_L1` is a boolean fixed column (=1 only at row 0, `main.pil:19`), `mirror = (1-segment_l1) x generated` exactly. The tool's own cofactor search prints "IMPLIED BY generated #4: this clause is (1 - pre(0)) * that constraint". No second discrepancy.

**2. `b_src_c` is byte-for-byte the mirror's expansion (proved).** `main.pil:350` and `:355`: `b_src_c = 1 - b_src_mem - b_src_imm - b_src_ind - b_src_reg`, identical in both config branches (unlike `a_src_c`). I confirmed these lines directly.

**3. The weakening cannot cost soundness (traced).** `sourceCCopyBetween` enters the proof as a **hypothesis** (`transitionBetween` -> `source_c_copy_lanes_of_between`, `StepStrongControlStore.lean:383`) — weakening a hypothesis is the safe direction. That consumer takes `h_segment : current.core.segment_l1 = 0` explicitly, exactly the region where mirror and generated are identical. The proof never reads the dropped `segment_l1 = 1` half. I verified both facts in the Lean source.

**4. The dropped half is a genuine interface limit.** At the boundary `b` is a public air-value (the boot/cross-segment datum, `main.pil:75`), and `Component.transition : Input -> Input -> Prop` has no air-value accessor — so the mirror *cannot* state it. The docstring's "no public-input surface" is real, not an excuse.

## Recommendation: (b) with a nuance

Declare #4/#10 as #103/#76 scope, but **distinct** from the Main #0/#19/#20/#21/#38 cluster. Those are modelled nowhere; #4/#10 are faithfully modelled across the entire in-scope `segment_l1 = 0` region, with only the out-of-scope boundary term dropped for a real reason. So the tool should read "cofactor `(1 - SEGMENT_L1)` over a boolean fixed column, remainder is #103/#76" as in-scope-coverage-plus-boundary-delegation, not a bare gap.

No proof change is needed for single-segment `root_soundness`. Closing the boundary would first require a public-input surface on the transition interface.

**Also re-flagged:** the a-side #3/#9 (`a_src_c`) has *no* mirror at all — also soundness-safe and unused, but **unscoped**, unlike #4/#10.

## Note on next steps (not in this PR)

The recommendation implies a tool change — teach `mirror-roundtrip` to recognise the `(1-SEGMENT_L1)` cofactor as coverage-plus-delegation, which would move #4/#10 out of the gap count. I did **not** make that change: reclassifying a finding is exactly the move that needs your sign-off, and it overlaps PR #321's tool work. Your call whether to fold it into #321 or leave #4/#10 as a declared exclusion.

Findings doc: `tools/mirror-roundtrip/FINDING_main_copyc.md`.


codygunton:
Addressed the review in `a21d390a`. All three were real; the central conclusion (weaker mirror, implied under the boolean relation, read by soundness only at `segment_l1 = 0`) is unchanged.

- **High — "already-tracked #103/#76" was wrong.** Confirmed both are **closed** and memory-specific: #103 (cross-segment Mem seam, closed 2026-06-18; its body says the old spike "should not be used as current implementation guidance") and #76 (load memory floor, closed 2026-06-20). Neither tracks Main's `segment_previous_c` source-C constraint; `RegisterBoundary.lean:30` is about register/memory continuation. Reframed §4 as an **untracked cross-segment limitation** needing a real tracking decision, structurally like — but not the same as — the `#0/#19/#20/#21/#38` cluster (which turns out to be only comment-referenced, not tracked either). Recommendation now says "do not attribute to #103/#76; open/repoint a tracking record".
- **Medium — 10-monomial diff mislabeled.** Corrected: the raw canonical diff is `generated − mirror = L·B·(x − s)` (`B` = 5 terms × `(x−s)` = 2 → 10 monomials). `(L−L²)·B·(s−p)` is the *cofactor-identity residue* (20 monomials), a distinct polynomial. Restated the equality as **polynomial equality modulo the boolean relation `L² = L`** (`(1−L)·generated − mirror = (L²−L)·B·(s−p)`), not a pointwise "iff".
- **Medium — boundary equation missing the selector.** Corrected to `b_src_c·(b − segment_previous_c) = 0`, forcing `b = segment_previous_c` only when `b_src_c ≠ 0` (=1 under selector well-formedness). Same fix applied to the recommendation's step 3.

codygunton:

PR #326, opened 2026-08-04
#325: replace raw provider-table binders with evidence bundles across ALU/Binary, shift, and BinaryAdd equiv layers

### LINKED ISSUE #325: Wrapper-binder hygiene: derive provider-table evidence inside the ALU/Binary equiv_<op> wrappers
Salvage lead from the Aristotle spike (#324, item 2); verified still applicable on main `29d6bccc`.

## What

The ALU/Binary `equiv_<op>` compliance wrappers still take caller-supplied provider-table evidence as explicit binders — e.g. `equiv_AND` (`ZiskFv/Compliance/Wrappers/And.lean`) takes `providerTable`, `providerRow`, `h_component`, `h_table_spec`, `h_provider_row`, `h_match`, `h_input_r1_row`, `h_input_r2_row`. The public trust surface is already clean: `root_soundness`'s frozen binders carry zero provider evidence because the StepStrong layer derives all of this from FullEnsemble balance at the call sites (~30 derivation sites in `StepStrongAluArith.lean` alone). PR #294 already did the end-state for the Div family by deleting all 16 `Inputs_div` fields.

## Why (and why it's optional)

This is internal plumbing only — no trust-surface change, since the binders are discharged internally today. The win is hygiene: the derivation logic currently lives at every call site instead of once inside the wrapper, and the wrapper signatures misleadingly suggest caller-supplied trust that no caller actually supplies. The spike's Phase-2 seam (stack R1–R4, `DerivedRowFacts`/`SharedBundles` evidence packages) prototyped the pattern but is not portable as-is; a fresh implementation against current main is smaller than a port.

## Scope

- Move the provider-row derivation inside the ALU/Binary `equiv_<op>` wrappers (or replace the binder lists with the derived-bundle shape the Div wrappers already use), family by family.
- `AeneasBridgeTrust/BinaryRType.lean` carries the same internal plumbing.
- Gates: `lake build` + V1/V2 trust gates must stay green; `root_soundness` binder baselines must be byte-identical (this change must not touch the public surface).

## Non-goals

No new public premises, no trust-ledger changes, no port of the spike's stack code.

codygunton:
Resolves #325.

## What

Replaces the raw provider-table binder lists in the ALU/Binary, shift, and BinaryAdd `equiv_<op>` layers with three evidence bundles, mirroring the derived-bundle shape PR #294 established for Arith. All bundles live in `ZiskFv/AirsClean/BinaryFamily/Balance.lean`, next to the lowering lemmas they package:

- `StaticBinaryProvider` — pinned to `Binary.staticLookupComponent`; `facts` lowers to the four row-level facts (`Spec`, `core_every_row`, `StaticBinaryTableSpecFacts`, `StaticBinaryTableWfFacts`) the `equiv_<op>_of_static_row` cores consume.
- `ShiftStaticProvider` — pinned to `BinaryExtension.shiftStaticLookupComponent`; `facts` gives `WfFacts ∧ ShiftB0RangeSpecFact`.
- `BinaryAddProvider` — pinned to `BinaryAdd.component` (the ADD/ADDI `via_binaryadd` route); `facts` gives `ComponentSpecFacts`.

Per opcode: the `Compliance/Wrappers` lemma and the canonical `Equivalence` theorem take the bundle; the canonical theorem now genuinely delegates to the wrapper (several docstrings always claimed this while the proofs re-derived everything inline — And, Or, Xor, Andi, Ori, Xori, Sub); the dead `equiv_<op>_of_static_table_row` variants (definition-only, zero callers — verified per file by repo-wide grep before each deletion) are deleted; Dispatch/Construction callers build the bundle inline from the same five fields they already destructure. 28 wrapper files, 28 Equivalence files, 6 Dispatch files, 8 Construction files. Net: **−1,252 lines**.

## Why

- The table→row lowering block was copy-pasted up to 3× per opcode (~84 copies across wrapper + canonical + dead variant). It now lives once per provider family.
- Wrapper signatures no longer suggest caller-supplied trust that no caller supplies: the component pin, table spec, and row membership travel as one named bundle, exactly like the Arith witness bundles after #294.
- No trust-surface change: the frozen global binder baselines bind `env : OpEnvelope` by type only and are byte-identical; per-op axiom closures unchanged (V2 verifies closures, not binder text).

## Explicitly out of scope (deferred leg, needs owner sign-off — protected surface)

`OpEnvelope.lean` constructors still carry the raw five fields, and the `AeneasBridgeTrust/*` envelope builders mirror them; the Dispatch arms bridge between the two shapes with an inline anonymous constructor (definitionally free). Converting the envelope itself is the leg flagged in #325's scope comment.

## Review evidence and steering record

Work split: coordinator authored the bundles + AND pilot end-to-end, validated it compiled first-try, then fanned out four parallel agents (R-type, I-type, Add/W-forms, shifts) with the pilot diff as template and disjoint file ownership; coordinator then closed the BinaryAdd route the agents correctly refused to guess at.

- Each agent verified every edited wrapper with `lake env lean` (28/28 exit 0) before the coordinator's full builds.
- Agent-flagged anomalies, all resolved deliberately: (1) both `via_binaryadd` lemmas were skipped as unrepresentable by the two original bundles — resolved by adding `BinaryAddProvider` rather than forcing a wrong pin; (2) the shift agent found the six W-shift arms duplicated across two dispatcher theorems in `Dispatch/Remaining.lean` (12 call sites, not 6) and converted both copies; (3) variant wrapper bodies (mode-pin shape in Sub/Slt/Sltu/Addi/Slti/Sltiu, W-mode in Addw/Subw, immediate-decomposition in Addiw) were mapped onto `facts` components while preserving each file's existing core-call argument order.
- Reviewer pass (coordinator): leftover-binder grep across all converted layers found exactly one stale docstring (Subw, fixed); full-diff spot-checks (e.g. `Equivalence/Srliw.lean`) confirmed binder-substitution-only changes with conclusions and non-provider hypotheses untouched; the 28 remaining `h_component h_table_spec h_provider_row` occurrences in Dispatch are OpEnvelope case patterns, intentionally raw.
- `trust/generated/clean-integration-audit.md` is not regenerated: its generator input (`baseline-caller-burden.txt`) was retired before this PR, so its per-op "caller pins providerTable.component" lines are a historical snapshot of the pre-bundle shape. The live gate (V1 check 9, `check-clean-integration.py`) is structural and passes unchanged.

## Verification

- `lake build ZiskFv`: 9,144 jobs, green (only pre-existing `DivSpinRootSoundness` linter warnings).
- V1 `trust/scripts/check-all.sh`: **19/19 PASSED** (zisk submodule seeded at pinned `9ca88b5b`).
- V2 `trust/scripts/check-all-semantic.sh`: **19/19 PASSED** (all raw closures ⊆ {propext, Classical.choice, Quot.sound}; binder baselines byte-identical).

### DIFF
--- a/ZiskFv/AirsClean/BinaryFamily/Balance.lean
+++ b/ZiskFv/AirsClean/BinaryFamily/Balance.lean
@@ -240,6 +240,31 @@ theorem StaticBinaryProvider.facts (p : StaticBinaryProvider) :
   rw [ZiskFv.AirsClean.Binary.staticLookupComponent_spec] at h_component_spec
   exact ⟨h_component_spec.1, h_core, h_component_spec.2, h_wf⟩
 
+/-- Caller-facing bundle for a BinaryAdd provider row backing an op-bus
+    request (the `via_binaryadd` ADD/ADDI route): same shape as
+    `StaticBinaryProvider`, pinned to `BinaryAdd.component`. -/
+structure BinaryAddProvider where
+  table : Table FGL
+  row : Array FGL
+  h_component : table.component = ZiskFv.AirsClean.BinaryAdd.component
+  h_spec : table.Spec
+  h_row : row ∈ table.table
+
+/-- The Clean BinaryAdd row selected by the provider evidence. -/
+def BinaryAddProvider.rowInput (p : BinaryAddProvider) :=
+  ZiskFv.AirsClean.BinaryAdd.component.rowInput (p.table.environment p.row)
+
+/-- One-shot lowering of the BinaryAdd provider evidence to the component
+    spec facts consumed by the `equiv_<op>_of_binaryadd_row` cores. -/
+theorem BinaryAddProvider.facts (p : BinaryAddProvider) :
+    ZiskFv.AirsClean.BinaryAdd.ComponentSpecFacts p.rowInput := by
+  have h_component_spec :
+      ZiskFv.AirsClean.BinaryAdd.component.Spec
+        (p.table.environment p.row) := by
+    simpa [p.h_component] using p.h_spec p.row p.h_row
+  simpa [BinaryAddProvider.rowInput,
+    ZiskFv.AirsClean.BinaryAdd.component_spec] using h_component_spec
+
 /-- Row extraction for a BinaryExtension operation-bus provider interaction. -/
 theorem exists_binaryExtension_row_eval_of_interaction_mem
     {table : Table FGL}
--- a/ZiskFv/AirsClean/BinaryFamily/Balance.lean
+++ b/ZiskFv/AirsClean/BinaryFamily/Balance.lean
@@ -203,6 +203,43 @@ theorem staticBinary_spec_facts_of_table_spec
   rw [ZiskFv.AirsClean.Binary.staticLookupComponent_spec] at h_component_spec
   exact h_component_spec.2
 
+/-- Caller-facing bundle for a lookup-aware static Binary provider: the
+    provider table, the selected row, and the three table-level facts that the
+    ALU/Binary `equiv_<op>` wrappers previously took as separate binders. The
+    balance derivations (`main_request_*_provided`) produce exactly these
+    fields. -/
+structure StaticBinaryProvider where
+  table : Table FGL
+  row : Array FGL
+  h_component : table.component = ZiskFv.AirsClean.Binary.staticLookupComponent
+  h_spec : table.Spec
+  h_row : row ∈ table.table
+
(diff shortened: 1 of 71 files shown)


codygunton (commented):
Reviewed against origin/main at ad8c31fd. No findings. The three provider bundles preserve the prior component pins and row-level facts; canonical equivalence theorems delegate to their wrappers as claimed; deleted static-table-row variants have no remaining callers; and no binder or trust-surface regression was detected. Verification: lake build ZiskFv passed 9,152 jobs; V1 passed 19/19; V2 passed 19/19.

PR #21, opened 2026-05-11
Step 1.7a: shared discharge-bridge helper — packed-lane BitVec reconstruction

codygunton:
## Summary

Adds `ZiskFv/Equivalence/Bridge/StateBridge.lean::bv64_packed_eq_of_lanes` — the opcode-independent arithmetic step every per-shape *discharge bridge* needs to convert transpile-axiom lane equalities (`m.a_0 = lane_lo (state.xreg rs1)`, `m.a_1 = lane_hi (state.xreg rs1)`) into the packed BitVec form `r1_val = BitVec.ofNat 64 (m.a_0.val + m.a_1.val * 2^32)`.

```lean
theorem bv64_packed_eq_of_lanes
    {bv : BitVec 64} {a_lo a_hi : FGL}
    (h_lo : a_lo = lane_lo bv) (h_hi : a_hi = lane_hi bv) :
    bv = BitVec.ofNat 64 (a_lo.val + a_hi.val * 4294967296)
```

Pure arithmetic, no new axioms. Stacked on #20.

## Why this exists (audit per anti-laundering principle)

Each per-shape bridge (BinaryAdd already in #19, Binary/BinaryExtension/Arith/Mem/ControlFlow staged for Steps 2b-f) needs the same arithmetic step. Factoring it once:

- avoids redundant per-bridge derivation work,
- gives a single point of audit if the BitVec/lane arithmetic ever needs to change,
- compounds savings as the remaining 5 shapes (~50 opcodes) all consume it.

## Plan updates (in `/home/cody/.claude/plans/plan-to-completely-resolve-wild-lynx.md`)

Added a new "Step 1.7 — Shared discharge-bridge helpers" section documenting:

* **1.7a (LANDED, this PR):** `bv64_packed_eq_of_lanes`.
* **1.7b (planned):** `SailStateBridge.lean::state_xreg_eq_of_read_xreg` for the Sail-side `read_xreg ↔ state.xreg` connection. Tractable; left for follow-up where it's needed.
* **1.7c (no new work):** existing `Tactics/ALURTypeArchetype.lean::main_row_in_alu_rtype_mode` and `ALUITypeArchetype.lean` already factor Main mode predicates across opcodes.

Step 2's header now says "All Step 2 bridges MUST consume Step 1.7's `bv64_packed_eq_of_lanes`". Each Step 2b/c/d/e/f section's "Reuses" list explicitly references it.

## Test plan

- [x] `lake build ZiskFv.Equivalence.Bridge.StateBridge` passes.
[...]

PR #16, opened 2026-05-10
Phase A: OperationBus permutation soundness axioms + ADD pilot

codygunton:
## Summary

- Adds the 6-axiom OperationBus permutation-soundness layer for `OPERATION_BUS_ID = 5000`, mirroring `MemBridge.lean`'s role for the memory bus.
- Demonstrates end-to-end use via a Phase B.1 pilot refactor of `equiv_ADD` that derives `matches_entry` from the new axiom rather than accepting it as a caller hypothesis.
- First step of `docs/fv/plans/op-bus-and-global-compliance.md` and the resolution path in `/home/cody/.claude/plans/plan-to-completely-resolve-wild-lynx.md`.

## Commits

- `6f9e985` — Phase A: OperationBus permutation soundness axioms (6 axioms across 6 provider classes — BinaryAdd, Binary, BinaryExtension, ArithMul, ArithDiv, ArithDivSecondary).
- `9e5ddef` — Phase B.1 pilot: thread OpBus axiom through equiv_ADD (drops `r_binary` parameter and the `matches_entry` user hypothesis; both come from `op_bus_perm_sound_BinaryAdd`).

## What's added

* `ZiskFv/Airs/OperationBus/Bridge.lean` (NEW) — the 6 axioms, each parameterized on the disjunction of `m.op` literals characterizing the provider's opcode coverage.
* `ZiskFv/Airs/OperationBus.lean` — adds `opBus_row_Binary` and `opBus_row_BinaryExtension` structural projections (mirrors the existing `opBus_row_Main` and `opBus_row_BinaryAdd`). Cross-checked against `bus_emission_{Binary,BinaryExtension}_0` in the extracted `Buses.lean`.
* Trust scaffolding: `trust/allowed-axiom-files.txt`, `trust/baseline-axioms.txt` (80 → 86), `trust/scripts/check-floor.sh`, `docs/fv/trusted-base.md` (new class #11), `.github/CODEOWNERS`.
* `ZiskFv/Equivalence/Add.lean` (refactored) — Phase B.1 pilot. V2 baseline diff: `equiv_ADD` now formally depends on `op_bus_perm_sound_BinaryAdd`.

## Trust class

PLONK / logUp permutation-argument soundness on `OPERATION_BUS_ID = 5000`. Same scope as the existing memory-bus permutation axioms (`lookup_consumer_matches_provider_load`, `memalign_load_perm_sound`).

## Verification

[...]

PR #327, opened 2026-08-04
S11/#304: weld-aware mirror round-trip against the generated constraints

### LINKED ISSUE #304: Round-trip the hand-written mirrors against the generated constraints

codygunton:
Closes #304. Consolidates and supersedes #319 / #321 / #322 (all based on the now-merged `s11-roundtrip`); rebased onto current `main` after the weld fan-out (#296) and #303 (#312) landed.

Pairs every comparable generated constraint against the hand-written mirrors that restate it, per AIR, by canonical polynomial equality — now **weld-aware**, since #296's `Iff.rfl` welds are kernel-checked mirror↔generated bindings the earlier version couldn't see.

## Result

```
165/176 covered (139 matched, 15 out-of-root, 4 bool-typed, 7 weld-covered);
9 gap, 2 strengthening, 3 unbacked, 2 reclassification, 0 unreachable, 1 undeclared delegation, 0 scope failure
```

**The 9 gaps are all Main, and they are exactly the "three documented model-scope gaps" the weld work (#310) left as S11's job**, at constraint granularity:
- **a-side C-copy** — Main #3, #9 (`main.pil:385`); no mirror anywhere (`grep a_src_c ZiskFv/` → nothing).
- **segment boundary** — Main #0, #19, #20, #21, #38 (`main.pil:86/423/426/508`); cross-segment seam, **untracked** (the #103/#76 it was attributed to are closed and memory-scoped — see the finding doc).
- **weakened b-side** — Main #4, #10; mirrored by `sourceCCopyBetween` but only in the within-segment form (`FINDING_main_copyc.md`, soundness-safe, reviewed).

Gate still exits 1 on these 9 — by design.

## What changed since the pre-weld version (16 gaps → 9)

The weld fan-out merging to `main` moved the tree; three updates track it, none forcing a number:

1. **Weld-aware coverage** (`weld_parse.py` + `WELD_COVERED` class). Parses the six `*MirrorWeld.lean` files, records every `(AIR, N)` on the RHS of an `Iff.rfl`/`by rfl` weld as covered (Lean's kernel already proved `mirror ↔ constraint`). Strict: implication welds and non-rfl `Iff` welds are recognized and *not* counted; a `constraint_N` in a comment or a helper argument is not a weld; unrecognized shapes raise. **All 7 MemAlignWriteByte gaps → WELD_COVERED** (welded to the out-of-root `boolean_sel_high_4b` mirror, which the AirsClean-only search never saw). **All 9 Main gaps stay gaps** — Main's rfl-welded set `{1,2,7,8,13,14,15,16,17,22,28}` is disjoint from them.
2. **Weld-helper classification** (`survey._weld_helpers`) clears 7 new "scope failures". Mechanical rule: an unclassified Prop in a `*MirrorWeld.lean` file is weld-internal iff it binds no row record (the `ReadsOnly*` predicates) or is pinned to constraints by an `Iff.rfl` weld (`gen36`). A fabricated row-binding, unpinned Prop still fails (acceptance `UNCLASSIFIED_WELD_MIRROR`).
3. **Reconcile `lanes.py` with #310** (`lanes.weld_column_failures`). Cross-checks the tool's symbol-table-derived column map against #310's checked-in `trust/generated/weld-columns/*.txt` and fails on disagreement. **They agree exactly today** — all 11 files, every column, every declared total. Read-only consumption of `trust/`; keeps lanes.py's independent derivation (which also covers fixed/exposed/stage-2 lanes the weld-columns files don't).

## Two flagged nuances (class left unchanged, reported as context)

- `cyclicSuccessorTransitionRows_weld`'s RHS mixes the `delta_pc` strengthening with 8 real constraints — those 8 are weld-covered; the strengthening finding stands.
- MemAlign #16 (a reclassification) is also rfl-welded (`boot_pc_zero_weld`) — the weld restates the constraint, but not via the aliased clause the finding rests on, so the reclassification stands.

## Evidence

`acceptance.py`: 20 cases + the lanes-vs-#310 cross-check, exit 0. 3/3 hand findings reproduced, 13/13 mutations classified as predicted, 3/3 neutral controls, real `ZiskFv/`/`trust/` byte-identical (checksum). New cases: `WELD_MUTATED_AWAY` (delete a weld → constraint reverts to gap), `UNCLASSIFIED_WELD_MIRROR`, forged column-map disagreement. Two honest adaptations to the post-#296 tree are documented in the README: six mutations on now-weld-covered constraints assert `MATCHED→WELD_COVERED` (their discriminating assertions preserved), and the `RomBoolSpec`-unreachable hand finding — **resolved by #296** (its only consumers are the weld files) — was relocated to a `WELD_CONSUMER_REMOVED` mutation.

## Scope / trust

Reports only; nothing under `ZiskFv/`, `trust/`, or `nix/` (beyond the gate wiring) modified. No axiom, no TCB change. `tools/pilout-roundtrip/check.py` (#303) still passes.

## Open decision for the owner

Gate fails on the 9 real Main gaps. Either (a) hard-fail, or (b) a citation-bearing shrink-only baseline (`check-shrinkage.sh` idiom). And three follow-on tracking issues to spin off #304: the untracked cross-segment cluster, MemAlignWriteByte scope (now shown weld-covered, so this may dissolve), and the 2 unbacked + delegation.

## Not verified

`nix run .#test` not run end-to-end. `check_mirrors.py` (exit 1, by design), `acceptance.py` (exit 0), and `pilout-roundtrip/check.py` (exit 0) all run directly.


codygunton:
Follow-on tracking filed (per the approved closeout plan):

- **#328** — Model Main's public-input / segment-boundary air-value surface. Closes this PR's cross-segment cluster: the 9 Main gaps + 2 b-side strengthenings. Root cause is `FormalEnsemble FGL unit` (`FullEnsemble.lean:293`) — no public-input surface for these checks to attach to. This is the substantive resolution (weld + bind).
- **#329** — the residual 6 findings (3 unbacked, 2 reclassification, 1 undeclared delegation); fix-or-declare each.

Together #328 + #329 bring `check_mirrors.py` to `176/176, 0 failing` and green step 4/10. This PR ships hard-fail by design; merging reds `nix run .#test` until #328/#329 land — accepted.

Issue #332, opened 2026-08-05
MemAlign L1 modeled as an unpinned witness (preL1): fixed-column lane-class gap


codygunton:
Surfaced by the #329 residual disposition of the #304 mirror gate. **A real fidelity gap, currently soundness-inert.**

## What

MemAlign `constraint_16` (`mem_align.pil:121`, `MemAlign.L1 * pc`) reads the **fixed** column `MemAlign.L1` (`preprocessed (column := 0)`). But the MemAlign mirror models it as an **ordinary witness field** `Valid_MemAlign.preL1` / `MemAlignRow.preL1` — MemAlign declares no `fixedColumns` schema at all (unlike Main/Mem). This is pre-existing behaviour disclosed in `ZiskFv/AirsClean/MemAlignMirrorWeld.lean`'s module header (point 2, from #296).

Consequence: a prover controlling `preL1` can set it to `0` on the boot row and **evade `pc = 0`** — which a genuine fixed column would not permit. The weld theorems `boot_pc_zero_weld` / `component_spec_weld` are explicitly documented there as **not** certifying `constraint_16`'s fidelity.

## Why it's inert today (but still a gap)

No proof under `ZiskFv/Compliance/` consumes `boot_pc_zero` / `preL1` for any conclusion (grep-confirmed: `boot_pc_zero`/`preL1` appear only in MemAlign's own AIR modules and the weld, not in the `root_soundness` composition). So the gap does not currently affect `root_soundness` — but it is a genuine lane-class fidelity gap, and if a future proof relied on the MemAlign boot `pc = 0`, it would be unsound as modeled.

## The fix (not minimal)

Give MemAlign a genuine `fixedColumns` schema and pin `L1` to it (as Main does via `mainFixedColumns` + `eval_mainFixedColumns_segment_l1`), touching `MemAlign/{Row,Circuit,Bridge,Constraints,Soundness,Spec}.lean` and the weld bridges together. Not a local change.

## Interim status in the #304 gate

Declared in `tools/mirror-roundtrip/check_mirrors.py` as `DECLARED_EXCLUSIONS["memalign_l1_disclosed_gap"]`, explicitly labeled "a live, open residual, not resolved," re-verified live every run (it withdraws if MemAlign gains a `fixedColumns` schema — acceptance mutation `MEMALIGN_FIXED_COLUMNS_TEXT_ADDED` proves the finding resurfaces), and cited to this issue.

[...]

Issue #328, opened 2026-08-04
Cross-segment continuation soundness (out of current single-segment scope)


codygunton:
**Status: out of current scope. Posterity issue — resolve if/when the soundness scope expands from single-segment to multi-segment (continuations).** Surfaced by the #304 mirror round-trip (PR #327).

## What this is

`root_soundness` (`ZiskFv/Soundness.lean:111`) is **single-segment**: it proves each segment is
internally sound *given* its boundary, with the boundary carried as the named `bootSeed` premise
(tagged #115/#119). It does not prove that segments **chain** correctly — that segment N's exposed
output equals segment N+1's input.

The circuit enforces that chaining with 9 Main constraints, all reading `Extraction.Circuit.exposed`
air values — the segment-boundary data:

| generated | PIL | value | kind |
|---|---|---|---|
| `constraint_0` | main.pil:86 | `main_last_segment` booleanity | pure continuation (flag) |
| `constraint_19` | main.pil:423 | `segment_next_pc` | pure continuation (output boundary) |
| `constraint_20`/`_21` | main.pil:426 | `segment_last_c[0/1]` | pure continuation (output boundary) |
| `constraint_38` | main.pil:508 | `segment_initial_pc` | continuation (input boundary / start pin) |
| `constraint_4`/`_10` | main.pil:386 | b-side source-C copy | MIXED: within-segment carry (in scope, already modeled by `sourceCCopyBetween`) + SEGMENT_L1-gated boundary term (continuation) |
| `constraint_3`/`_9` | main.pil:385 | a-side source-C copy | MIXED: within-segment carry (in scope but consumed by no proof) + boundary term (continuation) |

These are unmodeled because the ensemble is `FormalEnsemble FGL unit` (`FullEnsemble.lean:293`) —
`PublicIO = unit`, never plumbed to a component `transition`. Note Mem handles its analogous segment
constraints the same weld-only way (`ExtractedMemTrace`/`traceExposedValue` live entirely inside
`MemMirrorWeld.lean`, consumed by nothing in soundness), so this is a **system-wide** gap, not
Main-specific.

## Why it's out of current scope (not a bug)

In the single-segment theorem the boundary is an explicit premise (`bootSeed`), and the output-
boundary constraints (segment_last_c, segment_next_pc) have no consumer. Omitting them is
soundness-safe *for the single-segment claim*. The gap is only load-bearing when segments are
[...]

PR #18, opened 2026-05-10
trust gate: anti-laundering safeguards + canonical promise-discharge vocabulary

codygunton:
## Summary

- Adds two new V1 trust-gate checks (#7 and #8) that catch refactors that LOOK like progress on *promise discharge* but actually just rearrange the trust surface.
- Adds a Glossary section to `docs/fv/known-gaps.md` formally defining the project's canonical vocabulary.
- Updates `CLAUDE.md` to require these terms throughout PR titles, bodies, commit messages, and sub-agent prompts.

## V1 check #7 — hypothesis-count anti-laundering

`trust/scripts/check-hypothesis-count.sh` reads `trust/baseline-hypothesis-count.txt`:

```
ZiskFv.Equivalence.Add.equiv_ADD total=45 hypothesis=30
…
```

Per-theorem counts must match exactly. Reductions are allowed (refresh baseline alongside refactor). Growth fails. Catches \"split one *promise hypothesis* into N smaller ones\".

## V1 check #8 — caller-burden ledger

`trust/scripts/check-caller-burden.sh` reads `trust/baseline-caller-burden.txt`:

```
ZiskFv.Equivalence.Add.equiv_ADD 014 h_match_clo [match] m.c_0 r_main = …
…
```

One line per binder, with name, position, category, and type snippet. Catches refactors that hold the count steady but reshape the trust.

Categories: `validator | state | entry | range | match | bridge | bus_shape | transpile | byte_chain | loose | row | instance | other`.

## Canonical vocabulary

`docs/fv/known-gaps.md` Glossary defines:

- **promise hypothesis** — caller-supplied parameter on a canonical `equiv_<OP>` theorem that the proof body substitutes without deriving.
- **promise discharge** — replacing a *promise hypothesis* with a derivation from the *trust ledger*. Real *promise discharge* reduces the trust surface; renaming/splitting is laundering, not discharge.
- **discharge bridge** — Lean file under `ZiskFv/Equivalence/Bridge/<Shape>.lean` exposing a uniform discharge API.
- **trust ledger** — the 87 axioms in `trust/baseline-axioms.txt`.
- **caller-burden ledger** — `trust/baseline-caller-burden.txt`.
- **anti-laundering metric** — the pair of V1 #7 + #8 gates.
- **constructibility (separate gap)** — whether `Valid_<AIR>` is actually instantiable from a real ZisK trace.

[...]

PR #333, opened 2026-08-05
#329: close the mirror-gate residuals (fix MemAlign L1; declare the rest) — gate now green

### LINKED ISSUE #329: #304 residual mirror findings: unbacked, reclassification, delegation
The non-cross-segment failing findings from the #304 mirror round-trip (PR #327). Six items; each needs a fix-or-declare disposition to bring the gate to exit 0. The cross-segment cluster (9 gaps + 2 strengthenings) is #328; this issue is everything else.

## Findings

**3 unbacked** — a mirror equation over a field the AIR has no comparable lane for:
- MemAlign `delta_pc` (`cyclicSuccessorTransitionRows`' 9th conjunct) — its only generated source is inside the *excluded* challenge-mixed `constraint_36`. Likely a cited scope note, not a fix.
- 2 Main — uninvestigated; need a scoping pass. May share #328's exposed surface (if so, fold there).

**2 reclassification** — right algebra, wrong lane KIND:
- MemAlign `L1` (`mem_align.pil:121`) — a fixed column modeled as the free field `preL1`; `MemAlign.component` declares no `fixedColumns`. Fix = model it as fixed, or declare as a cited lane-kind item.
- Main `SEGMENT_L1` alias (constraint_18) — declared lane-kind alias, benign; confirm and declare.

**1 undeclared delegation** — Mem `generatedTransition → ZiskFv.Airs.Mem.permutation_every_row` (`Mem/GeneratedTransition.lean:251`): delegate is in neither `survey.CLASSIFICATION` nor `survey.DELEGATED`, so its clauses are compared by nothing. Pre-existing tool bookkeeping — classify the delegate.

## Done when

Each finding is either modeled (canonical match / weld) or recorded as ONE cited category-level `DECLARED_EXCLUSIONS` entry in `tools/mirror-roundtrip/check_mirrors.py` (never per-index — the tool refuses that by design). Combined with #328, `check_mirrors.py` reaches `176/176 covered, 0 failing` and `nix run .#test` step 4/10 passes. Related: #304, #328.

codygunton:
Closes #329 and #332. Takes the #304 mirror gate from `174/176, 8 failing` to **`176/176, 0 failing`** — the gate now passes. One of the eight was a real fidelity gap and got a **real fix**, not a declaration.

## The 8 residuals, disposed honestly

| finding | disposition |
|---|---|
| **MemAlign #16** reclassification (`L1` fixed col modeled as witness `preL1`) | **FIXED** — real fidelity/soundness fix, #332 (below) |
| Mem undeclared delegation (`generatedTransition → permutation_every_row`) | **FIXED** — classified as challenge/stage-2 (std_sum/logUp accumulator); the category already excluded elsewhere |
| 2 strengthening (Main `sourceCCopyBetween` within-segment) | declared — cofactor-verified `(1−SEGMENT_L1)` weakening of the welded #4/#10; boundary is cross-segment (#328); soundness-safe direction |
| 3 unbacked (`delta_pc`, Main `addr0`/`addr2`) | declared — mechanically re-verified the fields are PIL `const expr`s / hint payloads, never committed as columns anywhere |
| Main #18 reclassification (`SEGMENT_L1` alias) | declared benign — Main has a real `fixedColumns` schema every row reads `segment_l1` from |

## The real fix — MemAlign L1 (#332)

`constraint_16` (`mem_align.pil:121`, `L1 * pc`) reads the **fixed** column `L1`, but MemAlign declared no `fixedColumns` and modeled `L1` as the witness `preL1` — a prover could set `preL1=0` at the boot row and **evade `pc=0`**. Mirroring Main's `mainFixedColumns` pattern: added `memAlignFixedColumns` (`L1 = [1,0,…]`), `memAlignFixedLayout` mapping effective slot 26 (`preL1`) to the fixed column, and `eval_memAlignFixedColumns_L1` proving every materialized row reads `preL1` from it. The component now declares `fixedColumns := some memAlignFixedColumns` (`rawWidth 31 → 30`), so the **AcceptedTrace enforces boot `pc=0` through a fixed column no prover controls** — the gap is genuinely closed, not disclosed-and-declared. The weld reads the fixed lane; MemAlign #16 is now a real fixed-lane match. (One ripple: `DivInputs.lean`'s rawWidth case `31 → 30`.)

## Anti-laundering (after the Track B review)

Every declared exclusion is **category-level with a cited source, re-verified live every run** (cofactor search / `lanes.lane_map` / textual pin — never a hardcoded name or per-index allowlist), and each is backed by a `acceptance.py` **withdrawal mutation** proving the finding re-surfaces if its basis is mutated away. `acceptance.py` is 23/23.

## Verification

- Full `lake build` green (**9146 jobs**), **0 new axioms** (shrinkage floor), no `sorry`/`native_decide`.
- `check_mirrors.py --quiet` → `176/176, 0 failing`, exit 0.
- `acceptance.py` → exit 0 (23 cases + lanes-vs-#310 cross-check).
- `check-weld-column-maps.py` → PASS (12 AIRs).
- `nix/test.nix` step 4/10 comment + README updated (the gate passes now).
- Not run: full `nix run .#test` end-to-end.

## Scope note

The MemAlign fix genuinely reduces a fidelity/soundness gap in that AIR. The declared exclusions are genuinely out-of-scope categories (cross-segment #328, or fields the pilout never commits), each tracked/cited — not silenced. `ZiskFv/` changes limited to MemAlign Circuit + the weld + a one-line Div rawWidth ripple.

### DIFF
--- a/ZiskFv/AirsClean/MemAlign/Circuit.lean
+++ b/ZiskFv/AirsClean/MemAlign/Circuit.lean
@@ -276,8 +276,121 @@ theorem cyclicSuccessorTransitionRows_memAlignIdleRow :
   norm_num [cyclicSuccessorTransitionRows, memAlignIdleRow, memAlignRowOf,
     memAlignValue0Of, memAlignValue1Of, memAlignLane]
 
+/-- ZisK instantiates both the MemAlign witness and fixed traces over this
+    physical domain (`zisk/pil/src/pil_helpers/traces.rs:328-331`,
+    `MemAlignFixed<F> = GenericTrace<MemAlignFixedRow<F>, 2097152, 0, 17>`). -/
+def memAlignFixedCapacity : Nat := 2097152
+
+/-- Map the 31 effective `MemAlignRow` slots to 30 raw witness slots plus the
+    one physical fixed column `L1`. The flattened `ProvableStruct` order puts
+    `preL1` at slot 26. -/
+private def memAlignFixedLayout (slot : Fin 31) : Sum (Fin 30) (Fin 1) :=
+  if h_preL1 : slot.val = 26 then
+    .inr ⟨0, by omega⟩
+  else
+    .inl ⟨slot.val - (if 26 < slot.val then 1 else 0), by
+      split <;> omega⟩
+
+/-- MemAlign's one fixed column for the currently modeled segment: `L1` is
+    `[1, 0, ...]` (`zisk/state-machines/mem/pil/mem_align.pil:120`). -/
+private def memAlignFixedValues (_slot : Fin 1) (row : Fin memAlignFixedCapacity) : FGL :=
+  if row.val = 0 then 1 else 0
+
+/-- Component-owned MemAlign fixed schema. `IndexedFixedColumns.fixedAt`
+    supplies physical-domain periodic access, while `Table.fixed_domain`
+    bounds every materialized MemAlign prefix by this capacity. -/
+def memAlignFixedColumns : IndexedFixedColumns FGL 30 where
+  capacity := memAlignFixedCapacity
+  capacity_pos := by decide
+  effectiveWidth := 31
+  fixedWidth := 1
+  layout := memAlignFixedLayout
+  values := memAlignFixedValues
+
+/-- The 30 raw witness cells of a MemAlign row, omitting the component-owned
+    `L1` fixed cell (`preL1`) at effective slot 26. -/
+def memAlignRawRow (row : MemAlignRow FGL) : Array FGL :=
+  #[row.addr, row.offset, row.width, row.wr, row.pc, row.reset,
+    row.sel_up_to_down, row.sel_down_to_up,
+    row.reg_0, row.reg_1, row.reg_2, row.reg_3, row.reg_4, row.reg_5, row.reg_6, row.reg_7,
+    row.sel_0, row.sel_1, row.step, row.sel_2, row.sel_3, row.sel_4, row.sel_5, row.sel_6,
+    row.sel_7, row.sel_prove, row.delta_addr, row.delta_pc, row.value_0, row.value_1]
+
+@[simp] theorem memAlignRawRow_size (row : MemAlignRow FGL) : (memAlignRawRow row).size = 30 := by
+  simp [memAlignRawRow]
+
+/- Every materialized MemAlign row reads `L1` from the component-owned fixed
+    schema. -/
+set_option maxRecDepth 4000 in
+theorem eval_memAlignFixedColumns_L1
+    (index : Nat) (data : ProverData FGL) (raw : Array FGL) :
+  (Eval.eval (Environment.fromArray (memAlignFixedColumns.materialize index raw) data)
+      (varFromOffset (F := FGL) MemAlignRow 0)).preL1 =
+      memAlignFixedColumns.fixedAt 0 index := by
+  rw [ProvableStruct.eval_eq_eval, ProvableStruct.varFromOffset_eq_varFromOffset]
+  unfold ProvableStruct.eval ProvableStruct.varFromOffset
+  simp only [instProvableStructMemAlignRow, ProvableStruct.eval.go,
+    ProvableStruct.varFromOffset.go]
+  rw [ProvableType.eval_varFromOffset]
+  simp only [explicit_provable_type, ProvableType.size,
+    IndexedFixedColumns.materialize, Array.getElem?_ofFn]
+  simp [IndexedFixedColumns.fixedAt, memAlignFixedColumns,
+    memAlignFixedLayout, memAlignFixedValues]
+
+/-- The component-owned `L1` fixed column marks the physical first MemAlign
+    row as the boot row (`mem_align.pil:120`, `col fixed L1 = [1,0...]`). -/
+theorem memAlignFixedColumns_L1_first : memAlignFixedColumns.fixedAt 0 0 = 1 := by
+  simp [IndexedFixedColumns.fixedAt, memAlignFixedColumns, memAlignFixedValues]
+
+/-- Away from the physical first row, `L1` is zero before the intrinsic
+    fixed-domain bound permits a periodic wrap. -/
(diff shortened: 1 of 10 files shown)


codygunton (commented):
Blocking finding: the new `mirror_unbacked_field_uncommitted` exclusion is too broad and recreates the laundering behavior this gate is meant to prevent.

In `tools/mirror-roundtrip/acceptance.py:1032-1048`, `LANELESS_CLAUSE_ADDED` injects the arbitrary constraint

```
row.delta_pc * (row.addr + row.wr * 12345) = 0
```

yet explicitly expects the gate to exit 0. `_declare_unbacked_uncommitted` excludes every clause whose unresolved fields have no pilout lane, regardless of the clause's assertions over committed fields. Thus a future or malicious mirror strengthening can be hidden merely by multiplying it by `delta_pc`. This is the same failure mode `check_mirrors.py:33-40` says the tool must detect.

Please keep the specifically reviewed existing residuals narrowly declared, while leaving newly introduced or structurally different UNBACKED clauses as failures.

codygunton:
## Blocker fixed — the UNBACKED exclusion is now a definitional-pin check (commit `800573ba`)

You were right, and it was exactly the laundering shape we guard against: `_declare_unbacked_uncommitted` excluded any clause *containing* an uncommitted field, so `strengthening * delta_pc` (asserting `strengthening = 0` wherever `delta_pc ≠ 0`) was silenced — and `LANELESS_CLAUSE_ADDED` wrongly encoded that as expected-pass.

**Fix:** the exclusion now additionally requires the clause be a **definitional pin** (`_unbacked_clause_is_definitional_pin`): every monomial of the canonical polynomial that touches an uncommitted ("unresolved") atom must be *that one atom at exponent 1* — no committed cofactor, no second uncommitted atom, no square. The three reviewed residuals pass (`delta_pc − (pc'−pc)`, `addr0 − a_offset_imm0`, `addr2 − …` all have the field at constant coefficient 1); `strengthening × delta_pc` does not (its monomials are `delta_pc·addr`, `delta_pc·wr` — two atoms each), so it stays a **failing** UNBACKED. This is the narrow declaration you asked for; mutated/laundered unbacked clauses remain failures.

**Verified independently** (I injected the laundered clause myself in a temp copy, not just via the test):
- `delta_pc * (addr + wr*999)` → failing `1 unbacked`, `excluded_by` empty, tool exit 1.
- `addr0 * (a_offset_imm0 + b_src_ind*12345)` → same.
- Real tree unchanged: `176/176, 0 failing`, the 3 genuine pins still `mirror_unbacked_field_uncommitted carries 3`.

**Tests:** `LANELESS_CLAUSE_ADDED` renamed `LANELESS_CLAUSE_LAUNDERED` and flipped to expect a failing UNBACKED; added `ADDR0_CLAUSE_LAUNDERED` so it's not delta_pc-specific. `acceptance.py` → 24/24. Exclusion citation updated to state the pin criterion. Diff is tools + README only.

Ready for re-review.

codygunton (commented):
Remaining blocker: `_unbacked_clause_is_definitional_pin` is vacuously true when all unresolved terms cancel.

`canonical(clause.expr)` removes zero-coefficient terms (`poly.py:78-80, 90-102`). Therefore a clause such as

```
row.delta_pc - row.delta_pc + row.addr = 0
```

still has `delta_pc` in `clause.laneless`, but its canonical polynomial contains only the committed `addr` term. The loop at `check_mirrors.py:1380-1386` sees no monomial touching an unresolved atom and returns `True`, so `_declare_unbacked_uncommitted` excludes this genuine strengthening and the gate stays green.

Please require the canonical polynomial to contain the expected unresolved atom(s), not merely require that any surviving unresolved monomial has the pin shape, and add a cancellation-laundering acceptance mutation. The direct multiplication case is otherwise fixed correctly.

codygunton:
Third blocker fixed in 96245ad7.

**The hole.** `_unbacked_clause_is_definitional_pin` tests the clause's *canonical* `poly.Poly` form. Uncommitted atoms that cancel during canonicalisation are gone before the per-monomial loop runs, so `delta_pc - delta_pc + addr = 0` → `addr = 0`: no monomial touches an unresolved atom, the loop is vacuously satisfied, and the old terminal `return True` excluded a committed-lane strengthening as if it were a genuine pin.

**The fix.** Require an uncommitted atom to *survive* canonicalisation — track `saw_uncommitted` in the loop and `return saw_uncommitted` instead of `True`. A clause that names a no-lane field but canonicalises to purely-committed content is not a pin and stays a failing UNBACKED.

**Verified directly, both directions, on the injected clause `delta_pc - delta_pc + addr*54321 = 0`:**
- pre-fix code: laundered it (declared as a pin, `176/176 OK`, exit 0)
- fixed code: surfaces it (`1 unbacked`, exit 1)

The three genuine pins (`delta_pc = pc'−pc`, `addr0 = a_offset_imm0`, `addr2 = store_offset + store_ind·a_0`) each keep a bare uncommitted monomial after canonicalisation, so they still exclude — real gate stays `176/176`, 0 failing, exit 0.

New acceptance mutation `LANELESS_CLAUSE_CANCELLED_LAUNDERED` locks this in: 25 cases total, 19/19 mutations classified exactly as predicted, real tree unchanged (sha256 match), exit 0.

codygunton (commented):
Re-review complete at `96245ad7`: no remaining findings.

The non-vacuity fix is correct: `_unbacked_clause_is_definitional_pin` now requires an unresolved atom to survive canonicalization, so the exact cancellation bypass (`delta_pc - delta_pc + committed strengthening = 0`) remains a failing, un-excluded UNBACKED. The new `LANELESS_CLAUSE_CANCELLED_LAUNDERED` mutation locks that behavior in.

Reviewer verification from the PR tip:

- `check_mirrors.py --quiet`: `176/176`, 0 failing, exit 0.
- `acceptance.py`: 25/25 cases, 19/19 mutations classified as predicted, exit 0.
- `git diff --check`: clean.

Approved in substance. GitHub does not permit this account to submit a formal approval on its own PR.

Issue #337, opened 2026-08-07
Extend Clean's channel-guarantee pattern to selector-gated (conditional) interactions


codygunton:
## Summary

Clean's channel-guarantee pattern — a typed `Channel` with a semantic `Guarantees`, consumers
declaring `channelsWithGuarantees` and using `Channel.pull`, providers discharging `Requirements` —
cannot currently express a **selector-gated (conditional) interaction**. ZisK's Main AIR emits every
register/memory access conditionally, so the pattern does not reach the component where it would do
the most work.

This issue records the blocker precisely and the proposed extension, so the pattern can be adopted
rather than worked around. Surfaced by the #330 Phase-0 spike.

## Where we are today

Nine channels are defined under `ZiskFv/Channels/`. Eight carry `Guarantees _msg _data := True`; the
exception is `ZiskFv/Channels/SpecifiedRanges.lean:32-47`, which carries a real predicate plus a
`memDistanceMessage_guarantees_iff` bridge lemma — the static-lookup idiom, matching Clean's own
`BytesTable` in `Clean/Examples/FibonacciWithChannels.lean:24-38`.

`grep -rn "channelsWithGuarantees" ZiskFv/` returns **zero hits**: no component in the tree registers
as consuming guarantees from any channel. Every emission uses `emit`/`emitted`, which sets
`assumeGuarantees := false`, so nothing receives a guarantee even where one exists.

Cross-AIR consistency is instead derived by hand in
`ZiskFv/AirsClean/FullEnsemble/Balance/MemBusRowBridges.lean` (~3,000 lines), which matches providers
by message equality plus component classification — e.g.
`ActiveMainRegisterBoundaryProviderRowMatchSpec` (line 2760), whose shape is exactly the conclusion of
Clean's `exists_push_of_pull`.

## The blocker

Three facts combine:

1. **`Requirements` fires at multiplicity 0.** `ChannelInteraction.Requirements`
   (`Clean/Circuit/Channel.lean:181`) is `eval env i.mult ≠ -1 → channel.Guarantees … `. It ignores
   `assumeGuarantees` entirely, and `0 ≠ -1`, so an *inactive* selector-gated push owes a guarantee
   about a message whose fields are unconstrained witness values. Unsatisfiable in general.

2. **A typed `Channel` cannot gate on multiplicity.** `Channel.Guarantees`
   (`Clean/Circuit/Channel.lean:12`) takes only `(message, data)`; `Channel.toRaw` derives both
[...]

Issue #330, opened 2026-08-05
Discharge InputsAgree: layered decomposition + the register-file simulation path


codygunton:
Forward-looking companion to #184 (which *audits/documents* the `InputsAgree` premises). This records **which layers of `InputsAgree` are dischargeable and how**, and the concrete proof path. Surfaced while analysing the #304 mirror-gate a-side operand-forwarding gaps (#3/#9).

## What `InputsAgree` assumes

`root_soundness` (`ZiskFv/Soundness.lean:111`) takes `inputsAgree` as a premise; per opcode it is `Inputs_<op>` (`ZiskFv/Compliance/TraceLevelExport/RowDataArithMem.lean:171` for ADD). The load-bearing fields:

```lean
h_a_lo_t : (mainOfTable …).a_0 i.val = Trusted.lane_lo ((sail_to_rv64 (binding i)).xreg (regidx_to_fin c.r1))
h_a_hi_t : … a_1 …                     -- operand a = Sail register r1
h_b_lo_t / h_b_hi_t : … b = Sail register r2
```

It **directly assumes the ZisK operand columns equal the Sail register-file values**, per step, via the `Trusted` representation bridge. This is the same shape for every opcode.

## The finding it fixes an error in

The a-side operand-forwarding constraints (`main.pil:385`, generated `constraint_3`/`_9`, `a_src_c * (a - previous_c) = 0`) are unmodelled in the #304 mirror. They are **not** continuation (as #328 currently frames them) and **not** vacuous (`AddSpin` sets `a_src_imm/mem/reg = false`, so `a_src_c = 1`). They are unmodelled-and-harmless because `root_soundness` never uses the operand-sourcing constraints at all — it assumes the operand *values* directly via `InputsAgree`. That assumption is what this issue is about.

## Layered decomposition of `a_0 = xreg[r1]` and dischargeability

1. **Operand multiplexing** (`a` = the selected source). **Fully dischargeable, no trust** — it is circuit arithmetic (the operand-sourcing constraints). The F-only slice reaches only the forwarding (`a_src_c`) and immediate sources; register/memory sources go through the MemBus (layer 2).
2. **Register/memory read** (selected value = ZisK register-file cell). **Dischargeable modulo the existing axiom** — it is a MemBus read (`mainARegPreInteraction`, `mem_op=3`); proving it is memory-bus soundness, which bottoms out in the ExtF permutation-soundness trust axiom the system already assumes. No *new* trust; not to zero.
[...]

Issue #320, opened 2026-07-30
root_soundness: ProgramRowsBinding and RawProgramDecode premises are unwitnessed (non-vacuity gap at the audit entrypoint)


codygunton:
## Summary

`root_soundness` (the raw-program endpoint introduced by the Phase 9 stack, #313–#318) has two premises that **nothing in the tree has ever satisfied**:

- `programBinding : RawProgramBinding.ProgramRowsBinding trace start addr rawProgram`
- `rawProgramDecodes : ∀ i, RawProgramDecode trace i (ziskStep i) start addr rawProgram`

together with the caller-supplied layout maps `start` / `addr` / `rawProgram`.

No theorem anywhere concludes `ProgramRowsBinding`, and there is no conversion lemma from the older `ProgramBinding`. Confirmed by search: every occurrence of `ProgramRowsBinding` is the definition, a docstring, or a hypothesis position.

## Why the existing witnesses do not cover this

The six concrete instantiations (`degenerate`, `addSpin` 2 rows, `addAddiSpin` 3, `divSpin` 4, `jalrSpin` 2, `sdLdSpin` 7) discharge every premise of what is now `stepSound_of_programDecodes`, including `bootSeed` and `hAvoidKnownBugs`. That evidence covers the premises the raw endpoint **shares** with the interior theorem (trace, `ziskStep`, `inputsAgree`, `bootSeed`, `hAvoidKnownBugs`) and carries over unchanged.

It does not reach the two premises above, because the implication runs in the wrong direction:

```
ProgramRowsBinding + RawProgramDecode  ⟹  ProgramDecode
```

The witnesses build `ProgramDecode` bundles by hand and satisfy the *consequent*. Establishing that the consequent is satisfiable says nothing about the antecedent. `sdLdProgramDecodes` never touches `ProgramRowsBinding`.

## Failure mode

If `ProgramRowsBinding` were unsatisfiable — for instance if its surjectivity clause (every physical ROM row must be a primary or successor row of some raw word) conflicted with the separation and monotonicity clauses at some `programLength` — then:

- all six witnesses still compile unchanged,
- `lake build` is green, V1 18/18, V2 19/19,
- and `root_soundness` is vacuously true.

Nothing in the tree detects it. `#print axioms` says nothing about inhabitation.

[...]

PR #338, opened 2026-08-11
#330 Phases 0-7: root_soundness takes InputsAgreeCore + SegmentPcChain + StepRowsAligned

### LINKED ISSUE #340: Isolate the mem_op=3 register partition and classify its counterparts
Sub-issue of #330. Scope: the ZisK-side structural step that must exist before the register telescope can be walked at all.

## What "done" means

At `mem_op = 3`, a Main memory-bus interaction's balanced counterpart is provably one of the two **register-side** branches — another Main row's register pre-load push, or `RegisterBoundary`'s boot/reload — with every memory-side provider excluded.

## Why this is a real step, not bookkeeping

Before this, the register branch of the counterpart machinery was only ever used *negatively*: `BootSegmentMemorySeed.lean:942` rules `RegisterBoundary` out as the counterpart for `as = 2` rows. Nothing walked it forward. This supplies the dual exclusions, so the branch becomes usable.

The four memory-side providers are excluded on their own `mem_op` shapes:

| provider | `mem_op` | needs |
|---|---|---|
| `MemAlignReadByte` | literal `1` | nothing |
| `MemAlignByte` | `1 + is_write` | booleanity of `is_write` |
| `MemAlign` | `wr + 1` | booleanity of `wr` |
| `Mem` | `wr + 1` | booleanity of `wr` |

## Deliverable

`ZiskFv/AirsClean/FullEnsemble/Balance/RegisterChainBridges.lean`:

- `not_activeMainMemAlignReadByteProviderRowMatchSpec_of_main_mem_op_three`
- `not_activeMainMemAlignByteProviderRowMatchSpec_of_main_mem_op_three`
- `not_activeMainMemAlignProviderRowMatchSpec_of_main_mem_op_three`
- `not_activeMainMutableMemProviderRowMatchSpec_of_main_mem_op_three`
- `activeMainRegisterProviderRowMatchSpec_of_main_mem_op_three` — the classification
- `memBusMessage_eq_of_eval_emitted_provider_msg_eq` — full message equality for the telescope

## Explicitly NOT in scope

Walking the chain to `RegisterBoundary.bootMessage`. That needs the register-step descent, which is not modelled — see the sibling range-slice issue.

Closed by PR #338.

### LINKED ISSUE #341: PC arm: state it as one boot equation + a Sail-internal retire law (restructuring, NOT a discharge)
Sub-issue of #330. Scope: the PC arm of `InputsAgree` — `h_pc_bridge`, which is the one field present in **all 63** `Inputs_<op>` structures.

## READ THIS FIRST — this is NOT a strength reduction, and must not be closed as one

Both converses are proved:

- `pcSeed_of_inputsAgree` — the old per-row bundle follows from the seed;
- `sailRetireChain_of_inputsAgree` — the old per-row bundle plus `StepRowsAligned` yields `retire`.

So over an `AcceptedZiskTrace`, given `StepRowsAligned`, the old bundle and `boot` + `retire` are **inter-derivable**. This is the same bargain `BootSegmentMemorySeed` (#185/#115) struck for memory, which the ledger records as "net-zero-to-marginally-stronger". `trust/trusted-base.md` says so in those words.

The genuine PC discharge is a separate issue (chaining `SailTrace`).

## What "done" means

`root_soundness` states the PC premise as **one** cross-machine equation plus two one-sided laws, instead of 63 per-row cross-machine assumptions:

```lean
inputsAgree : ∀ i, InputsAgreeCore ziskTrace sailTrace i (ziskStep i)
pcChain     : SegmentPcChain ziskTrace sailTrace ziskStep   -- boot + retire
rowsAligned : StepRowsAligned ziskTrace ziskStep …
```

`boot` is the only premise relating a committed ZisK column to a Sail register. `retire` names no `mainOfTable` / `nextPcMux` / `pc` column. `rowsAligned` names no Sail state.

## Why it is worth having anyway

1. **`stepSound_of_programDecodes` becomes a strong induction** on the step index rather than a per-`i` map. Row `0` is `boot`; every later row is derived from the *previous row's own* `StepSound`.
2. **It exposed a condition the assumed `succ` was hiding.** `JalrLoweringRows` permits a two-row unaligned lowering with `finish = start + 1`, and `StepSound`'s JALR arm is indexed by `finish`, so such a step writes `pc (j+2)` into Sail's `nextPC` while step `j+1`'s bridge is stated at `pc (j+1)`. Assuming `succ` let a trace satisfy the PC equation while its Sail step had in fact retired elsewhere. Now recorded as `ZISK-MODEL-GAP-JALR-EXPANSION-STEP-ROW-INDEX` in `trust/defects.md`.

## Deliverable

`ZiskFv/Compliance/TraceLevelExport/SailRetireChain.lean` — `nextPC_of_busEffect_ok`, `busEffect_ok_{nil,one,three}`, `stepChannelOutput_busEffect_ok`, `sailInstructionOf`, `stepSound_iff`, `stepChannelOutput_execRows`, `SailRetireChain`, `StepRowsAligned`, `SegmentPcChain`, `pcBridge_succ_of_stepSound`. Plus the `SegmentPcSeed` split (`InputsCore_<op>` + child) and all 7 witnesses.

Gates: `lake build` 9158 jobs, V1 19/19, V2 exit 0, no new axioms, no new trust markers. Three gate artifacts updated deliberately (`Audit.lean` snapshot, `baseline-strong-export-binders.txt`, the V2 degenerate probe).

Closed by PR #338.

codygunton:
Closes #340. Closes #341.

Plan: `docs/ai/plan/PLAN_INPUTSAGREE_330.md`. Issue: #330.

## What this changes

The 63 per-row `h_pc_bridge` assumptions leave the root statement, and the successor law that
replaced them stops naming any ZisK column:

```lean
inputsAgree : ∀ i, InputsAgreeCore ziskTrace sailTrace i (ziskStep i)   -- was InputsAgree
pcChain     : SegmentPcChain ziskTrace sailTrace ziskStep               -- boot + retire
rowsAligned : StepRowsAligned ziskTrace ziskStep …                      -- ZisK-side placement
```

Phases 5/6 did the first step (`InputsAgreeCore` + a 2-field `SegmentPcSeed`); Phase 7 turned
`stepSound_of_programDecodes` into a strong induction so `SegmentPcSeed.succ` could be **derived**.

## Read this first — neither step is a logical-strength reduction

Both directions are proved, so be precise:

- `pcSeed_of_inputsAgree` (Phase 5/6) and `sailRetireChain_of_inputsAgree` (Phase 7) prove the
  converses. Over an `AcceptedZiskTrace`, given `rowsAligned`, the old per-row bundle and
  `boot` + `retire` are **inter-derivable**.
- Earlier wording on this branch (commit `fdccff32`, the `SegmentPcSeed.lean` module docstring)
  said "a genuine reduction rather than a repackaging". That was wrong; `ad57f637` corrects it in
  the docstrings, in `trust/trusted-base.md`, and in the plan.

This is the same bargain `BootSegmentMemorySeed` (#185/#115) struck for memory, which the ledger
records as "net-zero-to-marginally-stronger". It is legitimate because it is labelled honestly.
**#330 stays open.** The 116 `h_a_*_t` / `h_b_*_t` register occurrences are untouched — they are
Phases 3-remainder and 4, gated on the Clean repoint.

## How Phases 5/6 are built

`Inputs_<op>` splits into `InputsCore_<op>` (everything except `h_pc_bridge`) plus a child that
`extends` it and carries `h_pc_bridge` alone. That was chosen over 63 hand-written reduced
structures, and over threading a seed parameter through ~150 declarations and 413 application
sites. Inheritance keeps every consumer byte-identical: the 119 read sites, the `RowData_<op>`
bundles, the four `StepStrong*` files and `EnvOf` all compile untouched.

`inputsAgree_of_pcSeed` rebuilds each row's `h_pc_bridge` from the seed, feeding
`h_pc_bridge_of_pcSeed` the row's own Sail-PC naming, which every op already carried in one of
three places: 45 ops name it as `h_input_pc`, the 7 signed M-ext ops carry it in
`promises.input_pc_eq`, and the 11 memory ops carry it as the first conjunct of
`<op>_state_assumptions`. **No arm adds a premise.**

## Phases in this PR

| Phase | Content |
|---|---|
| 0 / S0.1 | `Spike330.lean` — multiplicity-gated channel consistency. GO. Not in the build graph; evidence only. |
| 2 | All four memory-provider exclusions at `mem_op = 3`, isolating the register partition |
| 3 (partial) | Register-partition classification + full message equality for the telescope |
| 5 | `SegmentPcSeed`; the ZisK-side PC recurrence; the `h_pc_bridge` consumption adapter |
| 6 | `InputsAgreeCore`; the root binder change |
| 7 | The PC-arm restructure — see below |

## Phase 7 — `succ` stops being a premise

`stepSound_of_programDecodes` is now a **strong induction on the step index** instead of a per-`i`
map, and the root takes:

```lean
pcChain     : SegmentPcChain ziskTrace sailTrace ziskStep    -- was pcSeed
rowsAligned : StepRowsAligned ziskTrace ziskStep ...         -- new
```

`pcChain` carries `boot` and `retire` (the Sail state at `j + 1` takes its `PC` from what step `j`
retired into `nextPC`). Row `0`'s PC agreement is `boot`; every later row is derived from the
**previous row's own `StepSound`**.

New module `ZiskFv/Compliance/TraceLevelExport/SailRetireChain.lean`:

- `nextPC_of_busEffect_ok` — if the channel effect succeeds, its post-state's `nextPC` **is** the
  producer entry's `pc`. Needs only the execution bus's shape; nothing about the memory bus, the
  opcode, or the row. This is what collapses 63 arms into one lemma.
- `busEffect_ok_nil` / `_one` / `_three` + `stepChannelOutput_busEffect_ok` — the channel effect
  never faults. `MemBusMessage.toEntry` takes multiplicity and address space as explicit arguments
  and every bus passes numerals, so `bus_effect`'s "impossible under assumptions" exits are
  unreachable. Proved, not assumed — which is why `retire` needs no existential.
- `sailInstructionOf` / `sailStepResult` / `stepChannelOutput` / `stepProducerRow` — the 63 arms of
  `StepSound`, named. `stepSound_iff` and `stepChannelOutput_execRows` then close by
  `cases zs <;> rfl`: `execute_instruction` *is* the `do` block the 30 spelled-out arms use, and
  every arm's execution bus is the same two-entry literal.

### Read this before calling Phase 7 a discharge — it is not one

`sailRetireChain_of_inputsAgree` proves the converse: the old per-row `h_pc_bridge` bundle **plus
`rowsAligned`** yields `retire`. So the two are inter-derivable given `rowsAligned` — the same
situation Phase 5/6 was in. `trust/trusted-base.md` says this in those words.

What genuinely changes:

- **`boot` is now the only premise relating a committed ZisK column to a Sail register.** `retire`
  names no `mainOfTable`, `nextPcMux`, or `pc` column — it relates two Sail states through Sail's
  own step function. `rowsAligned` names no Sail state. Each is dischargeable by reasoning on one
  side only.
- **A condition the old `succ` was hiding is now visible.** `JalrLoweringRows` permits a two-row
  unaligned lowering with `finish = start + 1`, and `StepSound`'s JALR arm is indexed by `finish`,
  so such a step writes `pc (j + 2)` into Sail's `nextPC` while step `j + 1`'s bridge is stated at
  `pc (j + 1)`. Assuming `succ` let a trace satisfy the PC equation while its Sail step had in fact
  retired elsewhere; nothing checked the two agreed. `rowsAligned` is that check. It binds only at
  steps that have a successor, so a trailing unaligned JALR — the `jalrSpin` shape — discharges it
  vacuously.

**The real reduction, still open:** define `SailTrace` as the sequence Sail's own semantics
generates from an initial state, making `retire` definitional and leaving `boot` alone. That is a
protected-interface change touching all 63 arms.

## Verification

- Full `lake build` green, 9156 jobs.
- V1 19/19 (`trust/scripts/check-all.sh`); V2 (`trust/scripts/check-all-semantic.sh`) exits 0.
- The new ZisK-side declarations close over `[propext, Classical.choice, Quot.sound]`. The ones
  that mention `StepSound` close over the pre-existing Sail LHS set the endpoint already carried.
- `root_soundness`'s axiom set is unchanged. No new trust markers anywhere.
- All 7 accepted-trace witnesses upgraded and still instantiate, including
  `addFaithfulPaddedRawRootSoundness` and `memoryRawRootSoundness`. **None evaluates a Sail
  execution by hand** — they build `retire` through `sailRetireChain_of_inputsAgree` from the
  `InputsAgree` family they already prove. `addFaithful`, `memoryRaw` and the degenerate probe get
  vacuous ones, as their `bootSeed`s already are.

## Two gate artifacts updated deliberately

Both are the documented workflow for an intentional statement change, and both firing is their
purpose:

1. `ZiskFv/Audit.lean`'s `#guard_msgs` frozen binder snapshot.
2. `trust/generated/baseline-strong-export-binders.txt`, regenerated via
   `lake exe trust-gate print-strong-export-binders`. The diff is exactly binder 10
   `InputsAgree` → `InputsAgreeCore`, binder 11 `pcSeed` → `pcChain`, and the new binder 12
   `rowsAligned` — nothing else.

The V2 degenerate probe `root_soundness_instantiation_degenerate.lean` gained a vacuous `pcChain`
and `rowsAligned` to match its existing vacuous `seed`.

## Also corrected here

`Soundness.lean`'s docstring and `trust/trusted-base.md` both still claimed #320 non-vacuity was
unwitnessed. That is false. `addFaithfulPaddedRawRootSoundness` instantiates `root_soundness` on a
one-instruction execution with the proved `addFaithfulProgramRowsBinding` and the real
`addFaithfulRawProgramDecodes`, so its conclusion is not vacuous. `memoryRawRootSoundness` supplies
a second real binding on an empty execution. Both texts now say so.

## Steering record

- The first design was a hand-rolled `BootSegmentRegisterSeed` mirroring `BootSegmentMemorySeed`.
  Dropped: Clean already proves the general channel-guarantee transfer theorem and we used none
  of it.
- Owner chose the `Guarantees` rewire over extending the existing counterpart machinery.
- Phase 0 then found that route blocked by a **Clean** limitation, not by our code:
  `ChannelInteraction.Requirements` fires at every `mult ≠ -1` including `0`, so every inactive
  selector-gated push would owe a guarantee about unconstrained witness values. Filed as #337.
- A message-content guard was tested and rejected before writing Lean: `aRegPreMessageExpr`
  hardcodes `mem_op := 3` while carrying `a_src_reg` as its multiplicity, so an inactive push still
  presents `mem_op = 3`.
- Phase 5 was therefore taken via the `Component.transition` fallback rather than the execution-bus
  route.

## Also in this PR, after the Phase 7 commits

**Phase 4 groundwork** (`fd37ee67`) — `regs_preserved_of_busEffect_ok_three` and
`regs_write_of_busEffect_ok_three`, the register analogues of `nextPC_of_busEffect_ok`: every
register other than the c-side write target and `nextPC` comes through unchanged, and the
destination register holds exactly the byte lanes the write entry carried. Kernel-only axioms,
opcode-uniform, and they slot into the induction Phase 7 installed.

They discharge **none** of the 116 `h_a_*_t` / `h_b_*_t` fields, and the module records the
measurement that explains why: `bus_effect` returns `Prop × EStateM.Result`, and its `.1`
accumulates exactly `read_xreg (wrap_to_regidx entry.ptr) state = .ok (U64.toBV #v[byteAt entry 0,
…]) state` for every `mult = -1, as = 1` entry — which *is* `h_a_lo_t` / `h_a_hi_t` in byte-lane
form. But `state_effect_via_channels` is `(bus_effect …).2`, so `StepSound` compares only the
post-state and discards `.1` entirely. Nothing in the export ever looks at the read conditions.
That is the structural reason those fields are assumed rather than checked.

**Defect ledger** — `ZISK-MODEL-GAP-JALR-EXPANSION-STEP-ROW-INDEX`, a `modeling-gap` recording that
`rowsAligned` narrows the theorem for an unaligned two-row JALR that is not the last executed step.
A narrowed claim belongs in the claim boundary, not only in prose. No `DefectId` constructor is
added and no baseline is touched: `DefectId` / `NoKnownDefect` is the per-row envelope-exclusion
family, whereas this is a segment-level structural hypothesis already frozen as a root binder by
`Audit.lean` and `baseline-strong-export-binders.txt` — stronger visibility than a defect
predicate. The entry says so explicitly, so the choice is reviewable.

## Known follow-ons, now tracked as sub-issues of #330

- **#343** — chain `SailTrace`; the genuine PC-arm discharge, per above.
- **#342** — model the register-step descent (bus-102 range slice). This is a **strict
  prerequisite** for #330's register arm: `a_reg_prev_mem_step` is a free witness column nothing
  constrains, so balance admits register-access cycles disjoint from `RegisterBoundary.bootMessage`
  and the 116 `h_a_*_t` / `h_b_*_t` fields cannot be discharged from the current model at any
  amount of proof effort. Not unsoundness of `root_soundness` — those reads are exactly what it
  assumes — and not a ZisK defect; the real circuit rules the cycle out at `main.pil:333-335`.
- **Phases 3 remainder / 4** — the 116 register fields. Blocked on the Clean repoint: `pulledIf` /
  `pushedIf` are confirmed absent from the populated Clean, so `flake.nix`'s `clean-src` must move
  `8edf71f8` → `bf5e40ed` and 16 files / 40 definitions must migrate to the post-#398
  `ElaboratedCircuit` class.


codygunton:
## Review — #338

Honest and well-documented PR. One **must-fix** that the description does not mention: it silently blinds a trust gate. Plus two docstrings that contradict a theorem in the same file.

### Must fix

**1. Trust gate 17(b) is now blind. Verified empirically.**

`trust/scripts/check-rowdata-partition.sh` matches `^structure (Decode|Inputs)_(\w+)`. `InputsCore_<op>` does not match that regex. This PR moves every substantive field into `InputsCore_<op>`, so check (b) — "no `Inputs_<op>` declares `h_main_op` / `h_main_active`" — now inspects 63 wrappers that hold only `h_pc_bridge`.

I injected `h_main_op : True` into `InputsCore_sub` on a scratch worktree of `330-spike` and ran the gate:

```
trust-gate: RowData partition integrity holds — 63 Decode_<op> are sailTrace-free;
none of 63 Inputs_<op> carries a decode pin (h_main_op / h_main_active).
EXIT=0
```

The script's own parser-sabotage guard does not fire, because the 63 wrappers still exist.

Fix: match `(Decode|Inputs(?:Core)?)_`, and raise the sabotage guard to expect 63 `InputsCore_<op>` as well. The PR's own content is clean — the `sub` / `and` splits add exactly `h_pc_bridge` — so this is a gate repair, not a proof defect. But the gate is the thing that stops the next PR from putting a decode pin there.

**2. Two docstrings contradict `sailRetireChain_of_inputsAgree`.**

* `SailRetireChain.lean:31-41` — heading "Why this is a strength reduction and not another repackaging", and "The converse does not hold: `SailRetireChain` cannot be recovered from `succ`".
* `SailRetireChain.lean:760-762` (`SegmentPcChain`) — "This is not: `retire` cannot be recovered from the per-row `h_pc_bridge` fields".

`sailRetireChain_of_inputsAgree` at `SailRetireChain.lean:791` recovers exactly that, and its own docstring says "the docstrings must not claim one". The PR body, `trust/trusted-base.md`, and that third docstring are correct. Commit `ad57f637` fixed the Phase 5/6 wording and missed these two. An overclaim in a docstring on the export path is the failure mode the anti-laundering rule targets.

### Should fix

**3. `Soundness.lean`'s root docstring still describes the superseded binder.** Lines ~43-70 describe `pcSeed` with `boot` / `succ`; the Phase 7 paragraph is appended below it, so the file states two different premise sets. The NON-VACUITY note also lists `pcSeed` among the shared premises. The same drift runs across 65 sites: `Dispatcher.lean:262` and the 63 per-op docstrings say the dispatcher builds `h_pc_bridge` via `inputsAgree_of_pcSeed` / `h_pc_bridge_of_pcSeed`. The live path is `inputsAgree_of_pcBridge`, inside the induction.

**4. Dead surface.** These have no consumers on the branch:

* `inputsAgree_of_pcSeed` (referenced only from docstrings);
* the seven `*PcSeed` defs in the witness files (only `.boot` is reused, via `pcSeed_of_inputsAgree`);
* `memBusMessage_eq_of_eval_emitted_provider_msg_eq`;
* `activeMainRegisterProviderRowMatchSpec_of_main_mem_op_three` — the terminal of the Phase 2 exclusion chain.

The `nextPcMux` route is also off the live path. `pcBridge_succ_of_stepSound` reads `pc (j + 1)` straight off the execution bus via `aligned`; it never goes through `mainOfTable_pc_succ_eq_nextPcMux`, although the module header at `SailRetireChain.lean:28-29` says it does. Roughly half of `SegmentPcSeed.lean` (651 lines) and the Phase 2/3 half of `RegisterChainBridges.lean` (507 lines) are groundwork for a blocked phase. Keeping them is defensible; the module docstrings should then say Phase 5 is retained as converse-direction evidence only.

**5. `Spike330.lean` never compiles.** `lakefile.toml` sets `roots = ["ZiskFv"]`, so the root-level file is outside the build graph — V1 check 18 confirms it is not among the 718 reachable modules. `gated_consistent` is therefore unelaborated evidence. It carries no trust markers. #339 declares it superseded upstream but does not delete it; delete it there.

**6. Minor.** The `-- #100: …` comment above `h_pc_bridge` in `Inputs_sub` still says "these" and mentions a no-wrap bound. Only one field remains under it.

### What checks out

* The split adds exactly `h_pc_bridge`, and `inputsAgree_of_pcBridge` rebuilds it from fields already in the core (`h_input_pc`, `promises.input_pc_eq`, `h_opcode_assumptions.1`). No arm gains a premise.
* `stepSound_iff` and `stepChannelOutput_execRows` close by `cases zs <;> rfl`, so the 63-arm duplication of `StepSound` is kernel-pinned against drift.
* No new `axiom` / `sorry` / `native_decide` / `opaque` anywhere in the diff.
* `trust/defects.md`'s `ZISK-MODEL-GAP-JALR-EXPANSION-STEP-ROW-INDEX` and the ledger entry are accurate, including the "why it is not a regression" argument.
* V1 on `330-spike`: 18 of 19 pass. Check 14 fails only because the scratch worktree has no `zisk/` submodule — an environment artifact, not a PR defect.

### Coverage note, no action

`rowsAligned` narrows the theorem, and the only two-row JALR witness in the tree (`RawProgramBindingJalrExpansionNonvacuity`) is at `numInstructions = 0`. Nothing exercises `rowsAligned` against a real two-row lowering. The retirement condition is recorded in `defects.md`, so this is a coverage observation rather than a defect.

codygunton:
Thanks — this was a good catch on the gate, and it was the one that mattered. All items addressed in `0fc035b6`.

## Must-fix 1 — the blinded gate

You were right, and I reproduced your sabotage before and after. `check-rowdata-partition.sh` now matches `(Decode|InputsCore|Inputs)_`, applies the decode-pin check to both `Inputs_` and `InputsCore_` (the error names which one), and the parser-sabotage guard requires ≥ 63 of all three kinds.

Injecting `h_main_op : True` into `InputsCore_sub`:

```
trust-gate: RowData partition-integrity FAILED:
  - …RowDataAluShift.lean: structure InputsCore_sub declares field `h_main_op`
    (circuit-only decode pin belongs in Decode_sub)
EXIT=1
```

and clean once reverted. Your framing was the right one: the PR's own content was fine, but the gate is what stops the *next* PR.

## Must-fix 2 — the contradicting docstrings

Both corrected. The `SailRetireChain.lean` header no longer claims a strength reduction or that the converse fails; it now states the two premise sets are inter-derivable given `StepRowsAligned`, matching the PR body, the ledger, and `sailRetireChain_of_inputsAgree`'s own docstring. `SegmentPcChain`'s docstring likewise.

You also caught that the header claimed `mainOfTable_pc_succ_eq_nextPcMux` is on the live path. It is not — `pcBridge_succ_of_stepSound` reads `pc (j + 1)` straight off the execution bus via `rowsAligned`. The header now describes the actual chain and says the `nextPcMux` route belongs to the retained seed evidence.

## Should-fix 3 — premise drift

`Soundness.lean` described two different premise sets. Replaced with one accurate description of `pcChain` + `rowsAligned`, naming which premise is cross-machine (only `boot`) and pointing at the JALR defect entry; the NON-VACUITY note updated. The 63 per-op docstrings and `Dispatcher.lean`'s `InputsAgreeCore` note now name `inputsAgree_of_pcBridge` / `h_pc_bridge_of_pcBridge`.

## Should-fix 4 — dead surface

Labelled rather than removed, with the reasoning stated in-file. `SegmentPcSeed.lean`'s header now separates what is load-bearing from what is **converse-direction evidence only** — deleting `pcSeed_of_inputsAgree` and friends would delete the proof that Phase 5/6 was a restructuring, which is exactly the claim the ledger rests on. `RegisterChainBridges.lean` says its Phase 2/3 half has no consumer yet and is blocked on #342.

## Should-fix 5 and 6

`Spike330.lean` deleted on #339, as you suggested. The `#100` comment no longer says "these" or mentions an absent no-wrap bound.

## Coverage note

Agreed and not addressed: nothing exercises `rowsAligned` against a real two-row JALR lowering. Happy to open an issue if you'd rather have it tracked than left as a `defects.md` retirement condition.

`lake build` green, 9156 jobs. V1 19/19 with the repaired gate.

codygunton:
Both correct. Fixed in `b9b50aba`, and the number re-derived from this head rather than restated.

## 1 — The V1 count

You are right, and the provenance matters more than the digit. Counted from each head directly:

```
$ git show 0fc035b6:trust/scripts/check-all.sh | grep -cE '^run '   # this PR
19
$ git show 9d1d61b0:trust/scripts/check-all.sh | grep -cE '^run '   # #339
20
```

**#338 is 19/19.** The channel-declaration snapshot is check 20 and it arrives in #339, so it cannot
be in this PR's count. This PR's body already said 19/19; the wrong number was in my summary
message, where I carried #339's total across. That is exactly the error you name — a gate result
quoted for a tree it did not run on — so it is worth being explicit that no artifact here was
affected, only my report of it.

Going the other way, #339's body understated *its* gate: it claimed 19/19 while adding check 20. I
corrected it to 20/20 with a note naming the base's 19/19, so the two PRs no longer report the same
total for different suites.

## 2 — The 63 per-op docstrings

Also right, and the same class as must-fix 2: one clause naming the retired premise while the rest
of the docstring describes the live path.

Changed `two-premise `SegmentPcSeed`` to `two-premise `SegmentPcChain`` in:

| file | occurrences |
|---|---|
| `RowDataAluShift.lean` | 23 |
| `RowDataArithMem.lean` | 31 |
| `RowDataControl.lean` | 9 |
| **total** | **63** |

63 is exactly the per-op set. `grep` confirms no occurrence of the stale name remains in those
files, and the clause appears nowhere else in the tree.

"two-premise" survives the rename unchanged, which is worth stating rather than assuming:
`SegmentPcChain` has exactly two fields, `boot` and the `retire` it inherits from
`SailRetireChain` — so the count was never the wrong part, only the name.

One thing I did **not** change, so it is a decision and not an oversight: the clause still does not
mention `StepRowsAligned`, which `root_soundness` also takes. You scoped the fix to one word, and the
sentence's subject is the PC-arm structure the dispatcher builds `h_pc_bridge` from rather than an
enumeration of root binders. Say the word if you would rather it name both.

Docstring text only. The three modules rebuild clean; full `lake build` and V1 re-running on
`b9b50aba` before merge.

PR #339, opened 2026-08-11
#337: migrate to Clean post-#398 and re-pin clean-src to bf5e40ed

### LINKED ISSUE #337: Extend Clean's channel-guarantee pattern to selector-gated (conditional) interactions

codygunton:
Closes #337.

**Stacked on #338** — base is `330-spike`, so review that first. 13 commits.

Plan: `docs/ai/plan/PLAN_INPUTSAGREE_330.md`, section "Clean upgrade".

## What this does

Migrates all 24 Clean components to the post-#398 `ElaboratedCircuit` API and repoints
`flake.nix`'s `clean-src` from `8edf71f8` to fork `bf5e40ed` (`codygunton/clean`, branch
`port-zero-mult-gating`, still Lean 4.28.0).

`ElaboratedCircuit` became a class indexed by `main`. `name` / `main` / `exposedChannels` /
`channelsWithRequirements` moved to `FormalCircuitBase`; `localLength` / `output` /
`channelsLawful` are auto-derived. Each old `@[reducible] def X : ElaboratedCircuit … where` plus
its separate `{ X with … }` collapses into one declaration.

## Why — and note what it unblocks, which is not what we expected

The repoint was wanted for `pulledIf` / `pushedIf`, to carry guarantees on selector-gated
interactions. It delivers that, and more: **#337 is fixed upstream**.
`Channel.toRaw.Requirements` is now `mult ≠ -1 → mult ≠ 0 → Guarantees`, which is exactly the
multiplicity gating #330 Phase 0 found missing. The bespoke `gated` `RawChannel` in `Spike330.lean`
is superseded.

**It does not unblock Phase 3.** With Clean out of the way, the real blocker turned out to be an
extraction-fidelity gap on the #169/#19 axis — the register telescope's strict descent
(`main.pil:333-335`, bus 102, 24-bit) is not modelled, so `a_reg_prev_mem_step` is a free witness
column nothing constrains. That finding, a concrete two-cycle witness showing the gap is
load-bearing, and the costed route are written up in the plan and in the worktree's `STATUS.md`.
It is out of scope for this PR.

## Safety gate: the channel snapshot

A bad mechanical rewrite here can drop a `channelsWithRequirements` or `exposedChannels` entry and
still compile — which changes the ensemble's balance obligations silently. That is the
anti-laundering failure mode, so every touched file is gated by `scripts/chansnap.py` (committed),
which extracts and normalises every channel declaration and diffs it against the pre-migration
revision.

**Result: 70 → 69, two deltas, both investigated rather than waved through.**

**−3, shadowed duplicates.** `Binary.staticLookupCircuit`, `BinaryExtension.staticLookupCircuit`
and `.shiftStaticLookupCircuit` each restated an `exposedChannels` its base already had,
byte-identically — the same declaration written twice, one shadowing the other. The merged form
declares it once. No interface is lost.

**+2, `channelsWithGuarantees := []`** on `MemAlignByte.circuit` and `MemAlignReadByte.circuit`,
which `pull` from MemBus. Post-#398 `elaborate_circuit` *derives*
`channelsWithGuarantees := [MemBusChannel.toRaw]`; pre-#398 the field was hand-written and took its
`[]` default. `[]` is restored explicitly. This is the one place the new API's default disagrees
with the pre-migration declaration, so here is the argument in full:

- `ChannelsLawful ops []` (`Clean/Circuit/Operations.lean:1134`) requires
  `∀ i ∈ interactions ops, i.channel ∈ [] ∨ i.Guarantees env`. The left disjunct is false, so it
  holds **only because** `MemBusChannel.Guarantees := True` (`ZiskFv/Channels/MemoryBus.lean:105`).
- It is therefore **kernel-checked, not assumed** — if `Guarantees` ever stops being `True`, these
  two declarations stop compiling. They cannot silently weaken.
- Both circuits carry `channelsWithRequirements := [MemBusChannel.toRaw]`, so
  `Operations.lean:302`'s `channels ⊆ guarantees ++ requirements` is satisfied.
- Accepting the derived list instead would make `SoundEnsemble.addTable`'s
  `channelsWithGuarantees ⊆ finished` obligation false at the ensemble's existing table order.

**Tripwire recorded in the plan:** when Phase 3 gives `MemBusChannel` a real `Guarantees`, both
declarations will fail to compile and the `addTable` ordering will bind on `fullRv64imEnsemble`.
Both are in scope for that phase. Neither is visible until the change is made.

## Trust-boundary note — read before merging

This PR pins a Clean revision in which `Channel.toRaw.Requirements` changes from
`mult ≠ -1 → Guarantees …` to `mult ≠ -1 → mult ≠ 0 → Guarantees …`. zisk-fv is on the *proving*
side of `Requirements`, so **an obligation this project discharges became weaker.** That is a
trust-boundary movement arriving inside an API-port PR, so it is called out here rather than left
to the diff.

**What it costs, precisely: a stated obligation, not a derivable conclusion.** Clean transfers a
requirement to a guarantee at `Clean/Air/Balance.lean:206`, and applies that step only to the
counterpart produced by `exists_push_of_pull`, which concludes `b.mult ≠ 0 ∧ b.mult ≠ -1`
(`Balance.lean:155-156`). The old and new definitions agree on that whole range. The case the new
definition drops — `mult = 0` — is therefore never consumed, so nothing derivable from the stronger
form is lost. Today it is moot twice over anyway: every channel we declare sets
`Guarantees := True`, which makes both forms trivial.

**Two caveats, which are the actual reason this needs a decision rather than a nod:**

1. **The argument above is human-made, not machine-checked.** It comes from reading `Balance.lean`,
   not from a Lean theorem. Nothing enforces "our `Requirements` are only consumed via
   `exists_push_of_pull`". A gate for that was considered and declined — the machinery costs more
   than the risk given `Guarantees := True` — and that decision is recorded rather than implicit.
2. **It is contingent on upstream.** If a later Clean revision consumes `Requirements` at
   `mult = 0`, the reasoning above stops holding. This should be re-read at the next `clean-src`
   bump, not assumed to carry forward.

**Where it first bites:** #342, which gives a channel a non-trivial `Guarantees` for the first
time. There Main's three bus-102 emissions produce obligations of exactly the gated shape
`¬(sel = 1) → ¬(sel = 0) → Guarantees …`, vacuous under selector booleanity. Without the gating
those would be real 24-bit range obligations Main cannot discharge. So the change is load-bearing
for the next phase, not merely tidy.

All of the above is also recorded in `trust/trusted-base.md`, which is the durable home; this
section exists so the reviewer sees it before merging rather than after.

## Other upstream API deltas

- The fork's `exists_nonzero_push_of_pull` is gone; upstream's `exists_push_of_pull` carries the
  same strengthened conclusion with the disequalities swapped. 8 sites rename and swap; no zisk-fv
  statement changed.
- `Component.rowOffset` / `.rowInputVar` went `abbrev` → `def`, causing 11 failures across 8 files
  including two `whnf` heartbeat timeouts. Each fix only names what the tactic used to get free.
- `elaborate_circuit` cannot derive `elaborated` for five circuits (unsolved
  `ExplicitCircuit (gatedDualStepDeltaRangeLookup _)`, or eight-interaction timeouts). All five
  state the `localLength` / `output` the old code wrote by hand.
- `GeneralFormalCircuit.Soundness` / `.Completeness` now take `main` plus an
  `[ElaboratedCircuit … main]` instance. Main's four standalone wrapper theorems restate over
  `main`, backed by two named instances (the Clean idiom, cf. `Clean/Circomlib/Poseidon.lean:241`).

## Verification

On this branch's HEAD, after rebasing onto `330-spike`'s tip:

- `lake build` — **green, 9158 jobs** (started at 45 errors).
- V1 `trust/scripts/check-all.sh` — **ALL CHECKS PASSED**, 20/20 (this PR adds check 20, the channel-declaration snapshot; the base #338 is 19/19).
- V2 `trust/scripts/check-all-semantic.sh` — **exit 0**.
- Channel snapshot re-run against the rebased base: same 70 → 69, same two deltas.
- **No new trust markers** — zero additions of `axiom` / `opaque` / `sorry` / `native_decide` or
  equivalents in the `ZiskFv/` diff.
- All 69 `Spec` / `Assumptions` / `ProverSpec` / `ProverAssumptions` lines preserved exactly
  (multiset-verified). `ZiskFv/Soundness.lean`, `Completeness.lean` and `ZiskFv/Airs/**` untouched.

## The lock and its documentation, moved together

`flake.lock`'s `clean-src` is re-pinned `8edf71f8` → `bf5e40ed` (via
`nix flake lock --update-input clean-src`), together with every file that describes the pin —
because if the lock and `flake.nix` disagree, CI populates the old Clean against migrated code:

- `nix/README.md` — the pin, plus what the repoint adds.
- `trust/trusted-base.md` — the build-input record, with a new paragraph stating plainly that the
  repoint closes #337 and that the gating change **weakens** the obligation we discharge. That is
  upstream's choice, not a local one, and it is why the affected sites now close with
  `intro _; simp [<Channel>]` instead of `intro _; trivial`.
- `docs/clean-fork-divergences.md` — **D0 deleted**, per that file's own maintenance rule: the
  nonzero-balance helpers and `Clean.Fin` namespace hygiene landed upstream in the PR-398 lineage
  the fork is rebased onto. A dated note records the retirement; D1–D3 provenance moves to
  `port-zero-mult-gating @ bf5e40ed`.

No reference to `8edf71f8` remains in the tree. V1 20/20 after the change.

**Build provenance.** The 9158-job build ran in `.worktrees/clean-migration-330`, whose `build/` is
a directory of symlinks to every shared entry **except** `clean-lean`, which is a local checkout of
the fork at `bf5e40ed` — i.e. the build is against the **new** Clean, the one `flake.lock` now pins.
The *shared* `/home/cody/zisk-fv/build/clean-lean` is deliberately still the old populated tree, so
the other worktrees kept building; `.lake` is per-worktree, so no olean crossed between them. CI
populates from the lock.


codygunton:
## Review — #339

Sound migration. The upstream claims check out against the pinned revision. Four items below; none blocks the substance.

### Verified against `bf5e40ed`

* `Channel.toRaw.Requirements` is `mult ≠ -1 → mult ≠ 0 → Guarantees` (`Clean/Circuit/Channel.lean:38-41`). #337 is genuinely fixed upstream.
* `exists_push_of_pull` concludes `b.mult ≠ 0 ∧ b.mult ≠ -1` (`Clean/Air/Balance.lean:155-156`), which matches the "disequalities swapped" note.
* The `channelsWithGuarantees := []` argument holds. `ElaboratedCircuit.channelsLawful` (`Clean/Circuit/Basic.lean:248-254`) is an autoParam that must close `ChannelsLawful main channelsWithGuarantees`. So `[]` is the *stronger* declaration, and it is discharged by a tactic rather than by a default. The tripwire note in the plan is the right call.

### Fix

**1. The comment on both `[]` sites is inaccurate.** `MemAlignByte/Circuit.lean:114` and `MemAlignReadByte/Circuit.lean:82` say "which is what `channelsLawful` below proves". There is no `channelsLawful` below — the obligation is the autoParam on the `elaborated` field. The claim is true; the pointer is not. Reword so a reader can find where it is discharged.

**2. `scripts/chansnap.py` is not a gate.** Nothing runs it: no CI job, no entry in `trust/scripts/check-all.sh`, and its only references in the tree are inside its own docstring. Describing it as "the safety gate" overstates what lands. Either wire it into `trust/scripts/`, or say in the file header that it is one-shot review evidence for this PR.

### Please answer

**3. Where did the green build run?** The PR says `build/clean-lean` on this machine is still the old Clean and that `nix run .#populate` was deliberately not run, then reports `lake build` green at 9158 jobs against the migrated API. Those cannot both describe the same tree. A green build against the *old* Clean would be meaningless for this PR, so state which worktree and which populated Clean produced the number. I could not re-run it — populating the new Clean would swap the API under every worktree on the machine.

### Sign off explicitly

**4. The `mult ≠ 0` gating weakens an obligation zisk-fv discharges.** `trust/trusted-base.md` says so plainly, and the practical effect is nil today because `MemBusChannel.Guarantees := True`. It is still a trust-boundary movement landing inside a PR titled as a mechanical migration, so it deserves a named approval rather than riding along with the API port.

### Also

`Spike330.lean` is declared superseded by this repoint, but the PR does not delete it. It sits at the repo root outside `roots = ["ZiskFv"]`, so it never compiles. Delete it here.

codygunton:
Thanks — all four addressed. Items 1, 2 and the `Spike330.lean` deletion are in `c99ed8fb`. Items 3 and 4 are answers, below.

## 3 — Where the green build ran

Fair catch; the PR body was ambiguous and you were right to refuse to accept the number on it. Both statements were true of *different* trees, which the body did not say.

The build ran in `/home/cody/zisk-fv/.worktrees/clean-migration-330`. That worktree's `build/` is **not** the shared directory — it is a real directory of symlinks to every shared `/home/cody/zisk-fv/build/` entry **except** `clean-lean`, which is a local `git clone` of `codygunton/clean`:

```
$ cd .worktrees/clean-migration-330/build/clean-lean && git log --oneline -1
bf5e40ed Air.Flat: use structure-instance notation in the *_mk lemmas
```

So the 9158-job build is against the **new** Clean, exactly the one `flake.lock` now pins. The shared `/home/cody/zisk-fv/build/clean-lean` is still the old populated tree, untouched, which is why the other nine worktrees kept building throughout.

`.lake` is per-worktree, so no olean crossed between them. I have added this to the PR body so the provenance is not folklore.

## 4 — Explicit sign-off requested on the `mult ≠ 0` gating

Agreed that it deserves a named approval rather than riding along, so stating it plainly rather than leaving it implicit:

**What moves.** `Channel.toRaw.Requirements` goes from firing at every `mult ≠ -1` to firing only at `mult ≠ -1 ∧ mult ≠ 0`. zisk-fv discharges `Requirements`, so this **weakens an obligation we prove**. It is a trust-boundary movement.

**Why it is defensible.** It is upstream's definition, not a local edit — we are pinning a revision, not patching one. The obligation it removes is the one for interactions at multiplicity zero, i.e. *inactive* selector-gated emissions, which is precisely the unsound-to-require case #337 was opened about: a disabled push owing a guarantee about unconstrained witness values. Practically the effect today is nil, because `MemBusChannel.Guarantees := True` makes both forms trivial.

**Where it is recorded.** `trust/trusted-base.md` states it in those terms, including that the weakening is upstream's choice and why the affected sites now close with `intro _; simp [<Channel>]` instead of `intro _; trivial`.

**Where it will bite.** #342 gives a channel a non-trivial `Guarantees` for the first time. There the gating is load-bearing: Main's three bus-102 emissions produce obligations of exactly the gated shape `¬(sel = 1) → ¬(sel = 0) → …`, vacuous under selector booleanity. Without the gating those would be real range obligations Main cannot discharge. That is the change working as intended, and it is the first place the difference is observable.

This is your call to approve, not mine to assume — flagging it as a distinct decision from the API port.

## On the coverage point you raised in #338

Noted and not addressed: nothing exercises `rowsAligned` against a real two-row JALR lowering, since the only such witness is at `numInstructions = 0`. The retirement condition is in `defects.md`. Worth an issue if you want it tracked rather than left as a ledger note.

Issue #162, opened 2026-06-27
Prove the extracted RV64IM decoder in-build (raw instruction word → opcode), matching the spec


codygunton:
## Status

Resolved. PR #164 landed on `main` on 2026-06-28 (merge commit `70d0c38b`). The proof is on current `main`.

## The gap

#160 proved ZisK's lowerer in-build. That proof maps an opcode to its decode fields for all 63 ops. But it starts from an already-classified `Rv64imSingleRowOpcode`.

The byte decoder `extract_decode_rv64im_raw` maps a raw 32-bit word to an opcode. The project extracted that decoder to Lean, but it had no theorem. Only the in-harness CI job spot-tested it with `native_decide`, and that job was retired (it cannot cold-build ProofWidgets; see PR #161). So the bytes→opcode step was unproven and untested.

## What PR #164 proved

PR #164 proves, in-build, that ZisK's real extracted decoder classifies each supported RV64IM raw word to the correct opcode. This completes the chain bytes → opcode (this issue) → fields (#160), all from the real Rust.

The proof has three layers:

- `zisk_decoder_accepts_supported_shape` (`ZiskFv/Compliance/AeneasBridgeTrust/Decode.lean`): the extracted opcode-support check accepts every raw word in `SupportedDecodeShape`.
- `zisk_decoder_classifies_supported_shape` (same file): the core decoder `decode_32_core` returns a decoded record, and that record holds a supported RV64IM opcode. The per-family `*_classifies` lemmas in `Decode/Classify.lean` pin the exact opcode for each shape.
- `zisk_decoder_classifies_matches_sail` (`ZiskFv/Completeness/Rv64im/SailDecode/ZiskClassifyMatch.lean`): for every raw word that the Sail spec decodes to a supported instruction, `decode_32_core` classifies that word to exactly the same opcode.

`real_decoder_accepts_in_shape` discharges the completeness obligation `OutstandingZiskPredicates.decoderAcceptsInShape` (`ZiskFv/Completeness.lean`). It needs no `knownDecodeGap` exclusion, because `SupportedDecodeShape` already restricts the FENCE family to the supported subset.

## Trust boundary

The ZisK decode half is kernel-sound. The axiom closure of `zisk_decoder_accepts_supported_shape` and `zisk_decoder_classifies_supported_shape` is `[propext, Classical.choice, Quot.sound]`. It uses no `native_decide`, no `bv_decide`, and no `sorry`.

[...]

PR #345, opened 2026-08-12
#342: model the bus-102 register-step descent; derive it from balance; clear the last sorry

### LINKED ISSUE #342: Model the register-step descent (bus-102 range slice) — prerequisite for discharging the register fields
Sub-issue of #330, and a **strict prerequisite** for its register arm. This is #169/#19 range-fidelity work.

## The gap, with a witness

The register telescope needs strict **descent** to terminate at `RegisterBoundary.bootMessage`'s literal `(ts 0, value 0)`. Balance alone only says pulls and pushes pair up as a multiset, so the pairing can decompose into the boundary path **plus disjoint cycles**.

Cycles are constructible in our model. A Main row's register-pre push and its memory pull carry the **same** value lanes (`Main/Constraints.lean:360-364, 396-401`); the pull's timestamp is pinned (`1 + main_step * 4`, with `main_step = i` from `MainStepIndexFixedFacts.main_step_eq_index`), the push's is the free witness column `a_reg_prev_mem_step`, which **no constraint in our Main component touches**.

Take two rows `A`, `B` on one register. Set `a_reg_prev_mem_step_A := ts_B`, `a_reg_prev_mem_step_B := ts_A`, both with the same arbitrary value `v`. Then `A`'s pull matches `B`'s push and vice versa — a closed 2-cycle touching neither boot nor reload. The boundary self-pairs, because our `RegisterBoundary` omits `main.pil:447`'s range check so the reload timestamp is free and can be `0`. `BalancedChannels` holds; both rows read a value nothing ever wrote.

**Consequence:** #330's 116 `h_a_*_t` / `h_b_*_t` fields **cannot be discharged from the current model at any amount of proof effort.**

## What this does NOT mean

- It does **not** make `root_soundness` unsound. Those reads are exactly what `InputsAgree` assumes, so the theorem is unaffected. Only the *dischargeability* is blocked.
- It is **not** a ZisK defect. The real circuit rules the cycle out.

## The missing constraint

`main.pil:333-335`, on bus **102**, 24-bit (`MAX_RANGE = (1 << 24) - 1`):

```
range_check(a_mem_step     - a_reg_prev_mem_step     - 1, min: 0, max: MAX_RANGE);
range_check(b_mem_step     - b_reg_prev_mem_step     - 1, min: 0, max: MAX_RANGE);
range_check(store_mem_step - store_reg_prev_mem_step - 1, min: 0, max: MAX_RANGE);
```

The extractor captures the wiring (`LookupWiring.lean` `hint_Main_40_1`: `piop := "Range Check"`, `busId := "102"`; `constraint_Main_41` for the b/store twins) but these are **stage-2 challenge-mixing** constraints, which the F-only extracted slice omits and Clean's `Channel`/`Air.Balance` layer models. `SpecifiedRangesSlice` is composed, but it is bus **103**, the 16-bit table Mem uses.

Taking the descent as a premise instead would be laundering: it is the entire content of register-read soundness, not a side condition.

## Status — substantially built, on `330-phase3-bus102-wip` (NOT GREEN, do not merge)

Green:

- `RegisterStepRangeChannel` (bus 102, real `Guarantees`, `rangeTable24`) and `ZiskFv/AirsClean/RegisterStepRangeSlice.lean`, the bounded static provider.
- Main's three emissions (`aRegStepDistanceExpr` and siblings), three-way `exposedChannels_eq` split. No new columns — `a_src_reg`/`b_src_reg`/`store_reg` are already stage-1 columns 35/36/37.
- `mainWithRomAndMemBus_selector_bool` + `mainWithRomMemAndOpBus_soundness`: the new `Operations.Requirements` conjuncts have the #337-gated shape `¬(sel = 1) → ¬(sel = 0) → …` and are vacuous under selector booleanity. **Main proves no range fact — it consumes one.**
- `fullRv64imEnsemble` composition. **No ensemble reorder needed** — `addTable` accepted the provider with no `channelsWithGuarantees ⊆ finished` complaint.
- The whole classification tail: `Classification`, `CounterpartClassification`, all three `*ProviderMatch` files.
- `registerStepRangeRowsTable` + `_constraints` + `_interactionsWith_of_ne` + `_interactionsWith` in `ConcreteRowReductions.lean`.

## Remaining

1. **Finish `singleAddWitness_registerStepRangeChannel_balanced`** — the only `sorry` in the tree, isolated and confirmed by `#print axioms`. Reduced to two interaction computations mirroring `binaryAddRowsTable_interactionsWith_opBus`; the provider side is done, the Main side and the final `AbstractInteraction.eval` step remain. Then the length bound is `3 + 3 < GL_prime` and `balanceOf` is `3·1 + 3·(−1) = 0`.
2. **Six more witnesses** — `AddSpin`, `AddAddiSpin`, `AddFaithfulPadded`, `DivSpin`, `JalrSpin`, `SdLdSpin`. Recipe established on `SingleAdd`: insert `registerStepRangeRowsTable <distances>` at the matching ensemble index, bump the `tables.length` literal, extend each table-list `rcases` by one arm, add bullets in the constraints / transitions / cyclic-successor / component-mismatch families, hoist the five `_interactionsWith_*_nil` lemmas above their uses, extend Main's channel lists from two to three entries, and extend the `BalancedChannels` dispatch to six.
3. **Then the original Phase 3**: walk the telescope using the descent plus `main_step_eq_index` to rule out cycles.

**No `sorry` leaves this branch.** The single one present is labelled in its commit message and must be discharged before merge.

codygunton:
Progress on #342. **Does not close it** — see "What remains" below.

This is #169/#19 range-fidelity work: it models the register-step descent (`main.pil:333-335`,
bus 102, 24-bit) that the register telescope needs to terminate, and which the F-only extracted
slice omits.

## Why this is needed

`a_reg_prev_mem_step` is a free witness column that no constraint in our Main component touches.
A Main row's register-pre push and its memory pull carry the same value lanes; the pull's timestamp
is pinned, the push's is free. So two rows on one register can point at each other's timestamps
carrying a fabricated value — a closed 2-cycle touching neither boot nor reload, with
`BalancedChannels` intact. #342 has the concrete witness.

Consequence: #330's 116 `h_a_*_t` / `h_b_*_t` fields cannot be discharged from the current model at
any amount of proof effort. This is **not** unsoundness of `root_soundness` (those reads are exactly
what it assumes) and **not** a ZisK defect (the real circuit rules the cycle out). It is extraction
fidelity.

## What this PR proves

**The descent, from balance rather than assumption.**
`rangeTable24_spec_of_registerStepRange_pull`: a Main bus-102 pull carrying `value` forces
`rangeTable24.Spec value`, so the pulled distance
`<slot>_mem_step - <slot>_reg_prev_mem_step - 1` is a genuine 24-bit number and the predecessor
timestamp sits strictly below the row's own.

The route has no new premise:

- `exists_push_of_pull` yields a counterpart with `mult ≠ -1` and `mult ≠ 0`;
- every other component is *proved* silent on bus 102 — nine new nil lemmas complete that family
  in `Classification`;
- Main is excluded because it only ever emits at `-selector`;
- the surviving counterpart is the provider slice, whose `Spec` **is** the 24-bit membership fact,
  discharged by the witnesses with `decide`.

`#print axioms` on all three new theorems: `[propext, Classical.choice, Quot.sound]`.

## The `sorry` is gone

`singleAddWitness_registerStepRangeChannel_balanced` was the only `sorry` in the tree, and it is
discharged. **No `sorry` term remains anywhere under `ZiskFv/`.**

The blocker was that Main's side of the interaction list stayed an unreduced
`AbstractInteraction.eval`. Three pieces close it: a Main exposed-channel lemma, the seven ROM
columns the emissions read as one conjunction (so the 25-field `mk` pattern is written once), and
one reusable per-emission eval lemma that every witness then shares.

## All seven witnesses close the channel

The regression floor still instantiates. Active pulls and provider lists:

| witness | rows | pulls | active | provider |
|---|---|---|---|---|
| SingleAdd | 1 | 3 | 3 | `[0,0,0]` |
| AddSpin | 3 | 9 | 3 | `[0,0,0]` |
| AddFaithfulPadded | 2 | 6 | 6 | `[0,0,0,1,0,0]` |
| AddAddiSpin | 4 | 12 | 5 | `[0,0,0,1,1]` |
| JalrSpin | 4 | 12 | 4 | `[2,2,10,7]` |
| SdLd | 8 | 24 | 10 | `[2,1,1,1,1,14,5,2,3,22]` |
| DivSpin | 5 | 15 | 5 | `[2,6,5,2,10]` |

Every distance was read off by `decide` / `#eval` rather than derived on paper. That was not
fastidiousness: the first hand-derived guess for AddAddiSpin's ADDI row was wrong.

Two changes here are not bookkeeping:

- **SdLd carries real prover data.** That exposed that `registerStepRangeRowsTable_constraints` and
  `_interactionsWith` were both silently pinned to `emptyData` through the environment. Both are now
  stated for an arbitrary `ProverData`, with the old forms as one-line corollaries. The slice
  genuinely does not read `data`; this makes that explicit instead of accidental.
- **`flatMap_fourteen_of_middle_nil`** (AddAddiSpin) is arity-specific, so it and its
  element/hypothesis indices widen to fifteen.

## Trust boundary

**No new trust markers** — zero additions of `axiom` / `opaque` / `sorry` / `native_decide` or
equivalent in the `ZiskFv/` diff (mechanically checked, not asserted).

One generated baseline moves, deliberately. V1 check 20 — the channel-declaration snapshot added by
#339 — fired on the interface change. The delta is exactly what this PR must produce:

- Main's `channelsWithRequirements` gains `RegisterStepRangeChannel` (it consumes bus 102);
- Main's `exposedChannels` gains its three bus-102 emissions;
- the provider declares `channelsWithRequirements := [RegisterStepRangeChannel.toRaw]` and
  `exposedChannels := [pushed (registerStepMessage value)]`.

No declaration was dropped or weakened; total 69 → 71, the two the provider adds. Regenerated by the
command that check's own header documents.

## What remains — this PR does not close #342

1. **Main's two-valued multiplicity is still an explicit hypothesis** on
   `rangeTable24_spec_of_registerStepRange_pull`. It follows from selector booleanity, which Main's
   constraints already prove (`romBoolSpec_of_mainWithRomMemAndOpBus_constraints`). Threading it is
   mechanical, and until it is threaded the theorem's usefulness depends on a caller supplying it.
2. **The telescope walk itself** — turning the per-row bound into well-founded descent terminating
   at `RegisterBoundary.bootMessage`. That is what actually kills the cycle globally rather than
   row-locally.
3. Phase 4 of #330 (the 116 fields) sits behind (2).

`RegisterStepRangeProviderMatch` therefore has **no consumer yet**; it is imported from the
`Balance` aggregator so the build compiles and checks it. That import is not cosmetic — V1 check 18
rejected the module for being unreachable, and until it was added the descent lemma was only
building because I targeted it directly.

## Verification

On this branch's head:

- `lake build` — **green, 9160 jobs**
- V1 `trust/scripts/check-all.sh` — **ALL CHECKS PASSED, 20/20**
- V2 `trust/scripts/check-all-semantic.sh` — see comment below
- no `sorry` anywhere under `ZiskFv/`; new theorems' axiom closure is the three standard axioms

**Reviewer note on history:** the first six commits are pre-existing WIP from the Phase 3 spike and
carry `NOT GREEN` / `do not merge` in their subjects. Those markers describe those commits, not the
branch — the tip is green on all three gates. Squashing on merge makes them moot.


codygunton:
**V2 green.** All three gates now confirmed on `0e9236e9`, which is this PR's head:

| gate | result |
|---|---|
| `lake build` | **exit 0**, 9160 jobs |
| V1 `trust/scripts/check-all.sh` | **ALL CHECKS PASSED**, 20/20 |
| V2 `trust/scripts/check-all-semantic.sh` | **ALL CHECKS PASSED**, exit 0, 19/19 |

Each was run on a clean tree at that commit (the only working-tree entry is pre-existing dirt
inside the `zisk` submodule, untouched by this branch).

Worth stating explicitly for review, since V2 is the gate that would notice: **no axiom-closure
baseline moved.** The only generated file this PR changes is
`trust/generated/baseline-channel-declarations.txt`, and V2's per-theorem axiom-dep baseline,
strong-export closure, `root_soundness` binder walk and false-probe canary all pass unchanged. The
new bus-102 machinery adds proofs, not trust.

codygunton:
## Update — both gaps from the original body are now closed, plus the walk's ingredients

### 1. The assumed multiplicity fact is gone

`main_registerStepRange_mult_cases` derives it from
`romBoolSpec_of_componentWithRomMemAndOpBus_constraints`: every Main bus-102 emission is
`-<selector>`, the selectors are boolean by Main's own constraints, so a Main interaction on this
channel sits at `-1` or `0` and can never be the `mult ∉ {0,-1}` counterpart `exists_push_of_pull`
demands.

`rangeTable24_spec_of_registerStepRange_pull` now takes exactly `witness.BalancedChannels`,
`witness.Spec` and `witness.Constraints` — three properties every accepted trace has, with nothing
assumed about the register columns.

### 2. The cycle is ruled out

`registerChain_nodup_of_descent`: a chain of register accesses linked by the bus-102 descent visits
no timestamp twice. That is the counterexample this issue was opened about — the balanced register
partition can no longer decompose into the boundary path plus a disjoint cycle, since a cycle would
need a timestamp to strictly increase back to itself.

Two supporting results make that honest:

- `prev_val_lt_of_registerStepSpec` — the range fact is about *field* elements, and field
  subtraction wraps, so on its own it orders nothing: a fabricated `prev` above `memStep` satisfies
  it vacuously by wrapping. Given both below `2 ^ 32`, which real row timestamps are (`k + 4 * i`),
  the wrap cannot occur and it becomes a genuine `<`.
- The pairing direction was **read off the emissions, not assumed**: `RegisterBoundary` boot-*pulls*
  at timestamp 0 and reload-*pushes* (`RegisterBoundary.lean:125-126`); Main pushes
  `aRegPreMessage` at `a_reg_prev_mem_step` and pulls `aMemMessage` at `1 + main_step * 4`
  (`Main/Constraints.lean:359-403`). My first reading of this was backwards, and checking it changed
  the shape of the induction.

### 3. The descent at the level of concrete rows

`rangeTable24_spec_{a,b,c}RegStepDistance_of_active` — for a Main row of an accepted witness whose
`<slot>` selector is on, the distance it pulls is forced into `rangeTable24`. All three slots.

### 4. The counterpart step as a timestamp equality

`memBusMessage_timestamp_eq_of_entry_match` — matching entries carry the same timestamp. Stated on
the entry match rather than on `ActiveMainSelfMemProviderRowMatchSpec`, because that spec is a
six-way disjunction over which of Main's six memory-bus emissions supplied the read.

## What still blocks the end-to-end walk, precisely

Collapsing that six-way disjunction means ruling out the three *mem* branches as providers. A
provider needs `mult ∉ {0, -1}`, and the mem emissions carry `-(a_src_mem + a_src_reg)` and
friends. Booleanity alone does **not** suffice: it allows the sum to be `2`. Ruling that out needs
**source exclusivity** — at most one of `a_src_mem` / `a_src_reg` set.

I checked, and no such constraint exists in the Main AIR: `MainRomSourceGuard` is about immediate
operands matching the committed program immediates, and it lives in prover assumptions, not
constraints. Exclusivity would have to come from the ROM-word decode that pins the flags (the #162
decoder axis).

So the walk is not blocked on proof effort here — it is blocked on a decode-side fact whose
provenance has to be established rather than assumed. Asserting it inside this PR would be exactly
the laundering the project's rules forbid, so I have stopped at the boundary and named it.

All new theorems' axiom closure remains `[propext, Classical.choice, Quot.sound]`.

codygunton:
**Retraction — the "blocked on source exclusivity" note above is wrong.**

I wrote that collapsing the six-way Main-self disjunction needs source exclusivity (at most one of
`a_src_mem` / `a_src_reg`), that booleanity alone allows their sum to reach `2`, and that no such
constraint exists in the Main AIR so the walk was blocked on a decode-side fact.

The first and third statements are true; the conclusion does not follow. The `mem_op` value does
the work:

- `aMemOpExpr = a_src_mem + 3 * a_src_reg`, so `mem_op = 3` forces `(0, 1)` under booleanity
  **alone** — the sum cannot be `2` *and* satisfy `mem_op = 3`.
- That makes the a-mem emission's multiplicity `-(a_src_mem + a_src_reg) = -1`, and
  `exists_push_of_pull` requires the counterpart at `mult ≠ -1`. Contradiction.
- `bMemOpExpr = (b_src_mem + b_src_ind) + 3 * b_src_reg` and
  `cMemOpExpr = 2 * (store_mem + store_ind) + 3 * store_reg` collapse the same way.

So the six-way disjunction narrows to the three register-pre pushes using only booleanity and the
provider's own `mult ≠ -1`. No exclusivity fact, nothing from the decoder.

Landed as `not_activeMainSelf{A,B,C}MemProviderRowMatchSpec_of_main_mem_op_three` (mirroring the
existing `mem_op = 1` family) and `selfMemProvider_registerPre_timestamp_of_mem_op_three`, which
concludes that the counterpart carries the consumer's read timestamp in its
`<slot>_reg_prev_mem_step`.

Composed with the bus-102 descent on the provider row and `prev_val_lt_of_registerStepSpec`, each
counterpart step moves strictly later in the trace — which is what
`registerChain_nodup_of_descent` needs.

codygunton:
## Adversarial review

Content only — no builds, no trust-gate runs. Read against the pinned `zisk` submodule and the generated extraction.

**Verdict: the core is sound and the fidelity checks out. The weak part is the tail — the last four lemmas claim more in their prose than they prove, and one carries a wrong numeric bound.**

### What I verified as correct

- **Source fidelity.** `main.pil:333-335` reads `range_check(expression: <slot>_mem_step - <slot>_reg_prev_mem_step - 1, min: 0, max: MAX_RANGE, sel: <slot>_reg)`, with `MAX_RANGE = (1 << 24) - 1` (`main.pil:11`). The model matches: same expression, same per-slot selector, `rangeTable24.Spec t := t.val < 2^24`.
- **Bus id and offsets.** `constraint_Main_41` carries the literal `102`, the `+1` / `+2` mem-step offsets, and negated selector multiplicities. `main_step_to_mem_step` (`mem.pil:513-518`) is `1 + 4*step + offset`, so `1/2/3 + main_step * 4` is right.
- **Baseline delta is purely additive.** 69 → 71, no declaration dropped or weakened.
- **Main proves no range fact.** The three `Requirements` conjuncts are vacuous under selector booleanity. The range fact comes from the provider through balance.
- **The multiplicity premise is genuinely discharged** by `main_registerStepRange_mult_cases`, not assumed.
- **All seven witnesses carry real distances.** That is the real non-vacuity evidence, and it is the strongest part of the PR.

### Findings

**1. The `< 2 ^ 32` timestamp bound is wrong for the pinned config.** `prev_val_lt_of_registerStepSpec`, `registerChain_*_of_descent` and `registerRead_timestamp_lt_provider_access` all demand `t.val < 2 ^ 32`. The docstring justifies this as "which real row timestamps do, being `k + 4 * i`". `config.pil:1,3` set `MAIN_STEP_BITS = 36` and `MEM_STEP_BITS = 38`, so a real mem_step reaches ~2^38. The bound is satisfiable today only because the model pins the segment away (`mainFixedColumns_main_step_eq_index`), while the extracted `STEP` is `main_segment * 2^22 + row_index`. The proof needs only `d + prev + 1 < GL_prime`. Relax the bound to 2^40 or 2^63.

**2. `h_mem` is dead in `prev_val_lt_of_registerStepSpec`.** `h_small` needs only `h_d` and `h_prev`; the final `omega` needs only `h_val`. Every caller must supply the premise for nothing, and `registerRead_timestamp_lt_provider_access` inherits the dead premise as `h_bound_cur`.

**3. `registerRead_timestamp_lt_provider_access` composes nothing.** Its docstring and commit message say it "joins the three pieces" and name `selfMemProvider_registerPre_timestamp_of_mem_op_three` and `rangeTable24_spec_aRegStepDistance_of_active`. The proof calls neither. It takes the link and the descent as hypotheses, then applies one lemma after a `simpa`. That is a pass-through wrapper, which `AGENTS.md` forbids. Either compose the two named lemmas or downgrade the docstring to "value-level restatement".

**4. `registerChain_nodup_of_descent` does not rule out the #342 cycle.** It quantifies over an arbitrary `List FGL` and assumes both the chain relation and the bounds. Nothing builds such a list from a witness. Its docstring claims "the balanced register partition can no longer decompose into the boundary path plus a disjoint cycle". That claim is not proved anywhere, and the PR's own "What remains" item 2 says the telescope walk is missing. The prose and the remaining-work list contradict each other.

**5. `main.pil:447` is still unmodeled and unmentioned.** Issue #342 names two parts of the hole. This PR closes 333-335 only. `registerBoundary_table_interactionsWith_registerStepRange_nil` now *proves* `RegisterBoundary` silent on bus 102, and that lemma is load-bearing in `exists_registerStepRange_provider_of_pull` — so modelling 447 later forces a rework there. Say in "What remains" whether 447 is still needed. (My reading is that Main-side descent alone may already kill the cycle, because every Main pull-to-push step must then increase strictly. That argument is not in the PR.)

**6. `trust/trusted-base.md` was not updated.** Two places. Line ~636 still names `main.pil:447` and the #169/#19 axis as undelivered. Lines ~1110-1130 still say "every channel we declare sets `Guarantees := True`" and forecast #342 as the first non-trivial one. The first half was already false — bus-103 `SpecifiedRangesSliceChannel` carries `rangeTable16.Spec` — and the second half is now history. `AGENTS.md` requires the Lean declaration, the generated ledgers and `trusted-base.md` to move together. Only the generated baseline moved.

**7. Minor.** `not_activeMainSelfCMemProviderRowMatchSpec_of_main_mem_op_three` binds `_providerInteraction` and then uses it; the a-side and b-side twins use the plain name. And `mainWithRomAndMemBus_selector_bool` selects its three facts by a 20-deep `.2.2….1` chain, while `main_registerStepRange_mult_cases` gets the same three selectors from the named `romBoolSpec_of_componentWithRomMemAndOpBus_constraints`. Two routes to one fact; prefer the named one.

### On the anti-laundering axis

No violation. The model narrows (a real constraint is added), the seven witnesses prove the narrowing is not vacuous, no trust marker is added, and the PR states plainly that it does not close #342. The overclaiming is confined to docstrings and commit messages on items 3 and 4, not to theorem statements.

codygunton:
Thank you — finding 1 is a real bug and I would not have caught it. All seven addressed; five in
code (`935aa2d5`), two answered here.

## 1 — the `2 ^ 32` bound was wrong. Fixed.

Verified at source: `zisk/pil/config.pil:1,3` give `MAIN_STEP_BITS = 36` and
`MEM_STEP_BITS = MAIN_STEP_BITS + 2`, so a real memory timestamp reaches `2 ^ 38`. My bound sat
below the ceiling and held only because the model pins the segment away — on a full-length segment
the premise would have been unsatisfiable, and every downstream lemma would have been vacuous
exactly where it mattered.

Relaxed to `2 ^ 40`. The docstring now cites `config.pil` and records that the earlier bound was
wrong, so the next reader does not have to rediscover it.

## 2 — `h_mem` was dead. Removed.

Correct: `h_small` needs only `h_d` and `h_prev`, and the closing `omega` needs only `h_val`.
`memStep` is *derived*, not assumed. Gone from `prev_val_lt_of_registerStepSpec`, from the two
chain lemmas, and from `registerRead_timestamp_lt_provider_access`, which inherited it.

## 3 — the wrapper composed nothing. Now it does.

You are right that it was a pass-through with a docstring describing work it did not do. Rather
than downgrade the prose I made the theorem earn it: it now takes `BalancedChannels` / `Spec` /
`Constraints` plus the provider row, derives the descent through
`rangeTable24_spec_aRegStepDistance_of_active`, pulls `a_reg_prev_mem_step` out of the link via
`eval_aRegPreMessageExpr`, and only then applies the bridge. It still takes the *link* as a
hypothesis, because the collapse lemma returns a three-way disjunction and this discharges the
a-side; the docstring says so.

## 4 — the nodup docstring overclaimed. Rewritten.

Also correct, and it contradicted the PR's own "What remains". The docstring now states the scope
in as many words: it quantifies over an arbitrary list, assumes the chain relation and the bounds,
and therefore does **not** by itself establish that an accepted trace's register partition avoids a
disjoint cycle. It is the order-theoretic half; the walk that extracts the list is the remaining
work.

## 5 — `main.pil:447`. You are right that it was unmentioned, and right about the rework.

Recorded in `trusted-base.md` and in "What remains": bus 102 delivers `main.pil:333-335` only. The
`RegisterBoundary` reload timestamp stays free, and
`registerBoundary_table_interactionsWith_registerStepRange_nil` currently *proves* that component
silent on bus 102 — that lemma is load-bearing in `exists_registerStepRange_provider_of_pull`, so
modelling 447 later forces revisiting it and the provider case split. Flagged in both places.

On your reading that Main-side descent alone may already kill the cycle: I think that is right, and
it is worth stating why. Every Main→Main counterpart step now increases the timestamp strictly
(`selfMemProvider_registerPre_timestamp_of_mem_op_three` composed with the descent), so a cycle
consisting only of Main rows is impossible without 447. What 447 buys is the *boundary* end — a
free reload timestamp still lets the boundary self-pair, which is about termination at
`bootMessage`, not about Main-side cycles. I have not proved that decomposition, so I have written
it as your reading plus my agreement rather than as a result.

## 6 — the ledger. Both passages fixed.

`AGENTS.md` is explicit that the Lean, the generated ledgers and `trusted-base.md` move together,
and I moved only the generated baseline. Now corrected:

- the `main.pil:447` residual paragraph records what bus 102 delivers, what 447 still does not, and
  the rework note above;
- the "every channel we declare sets `Guarantees := True`" sentence is corrected. As you say it was
  *already* false — bus-103 `SpecifiedRangesSliceChannel` carries `rangeTable16.Spec` — and #342
  adds a second. I also noted that the `mult ≠ 0` gating argument never depended on `Guarantees`
  being trivial; it depends only on where Clean consumes `Requirements`.

## 7 — binder naming fixed; the two-routes point noted.

The C twin now binds `providerInteraction` plainly, matching a/b.

On `mainWithRomAndMemBus_selector_bool`'s 20-deep `.2.2….1` chain: agreed the named
`romBoolSpec_of_componentWithRomMemAndOpBus_constraints` route is better, and that is what
`main_registerStepRange_mult_cases` uses. I have **not** rewritten the existing lemma in this PR —
it is pre-existing, load-bearing for Main's soundness, and changing it is unrelated to bus 102.
Left as a follow-up rather than folded in silently.

## On the anti-laundering verdict

Noted, and I agree with where you drew the line: the overclaiming was in docstrings and commit
messages, not theorem statements. That is still the failure mode worth catching, since those are
what the next reader trusts.

Gates re-running on `935aa2d5`; I will post build/V1/V2 when they land.

codygunton:
**Gates green on `935aa2d5`**, the review-fix head:

| gate | result |
|---|---|
| `lake build` | **exit 0**, 9160 jobs |
| V1 `trust/scripts/check-all.sh` | **ALL CHECKS PASSED**, 20/20 |
| V2 `trust/scripts/check-all-semantic.sh` | **ALL CHECKS PASSED**, exit 0, 19/19 |

Run on a clean tree at that commit. The relaxed `2 ^ 40` bound and the recomposed
`registerRead_timestamp_lt_provider_access` both build; no axiom-closure baseline moved, and the
only generated file this PR touches is still the channel-declaration snapshot.

Issue #172, opened 2026-06-28
root_soundness pairs raw instruction bits with Sail instructions by hand


codygunton:
`root_soundness` does not use Sail's decoder. It never did, and it does not need to. This issue addresses a different gap: inside `root_soundness`, no theorem ties the committed instruction bits to the Sail instruction the proof executes. The pairing is done by hand.

## What the proof does today

`root_soundness` (`ZiskFv/Soundness.lean:111`) takes three kinds of premise: a raw program image (`rawProgram`), a production-lowering certificate (`programBinding`), and per-step raw-word evidence (`rawProgramDecodes`). Each `RawProgramDecode_<op>` structure carries an `hLine` field. That field names a specific `raw*` bit-layout.

Take an ADD step as an example. `RawProgramDecode_add.hLine` (`TraceLevelExport/RawProgramBindingCopyb.lean`) requires the committed word at this step's PC to equal `rawRType 0 (regidx_to_fin c.r2).val (regidx_to_fin c.r1).val 0 (regidx_to_fin c.rd).val 0x33`. Then `programDecode_of_rawProgramDecode` (`TraceLevelExport/RawProgramDecode.lean`) feeds that raw word through `extract_transpile_rv64im_raw`. That function is the Aeneas-translated Lean image of ZisK's Rust lowering path. The result is a `ProgramDecode`. The dispatcher then proves that the ZisK transition equals `execute_instruction (instruction.RTYPE (c.r2, c.r1, c.rd, rop.ADD))`.

No decoder runs during the proof. ZisK's Rust holds the real decoder (`decode_32_core`) and the real lowering (`lower_rv64im_single_row_input`). Aeneas translates that source offline into `trust/aeneas/ProductionM2.lean`. CI verifies that the checked-in file is byte-identical against a fresh extraction. In Lean the result is an ordinary pure function. The proofs apply it to a symbolic word.

`execute_instruction` (`SailSpec/Auxiliaries.lean:940`) sets `nextPC` and calls `LeanRV64D.Functions.execute`. Sail's `execute` (`InstsEnd.lean:69929`) takes the `instruction` value as data and dispatches on the constructor. It does not decode.

## What we assume

[...]

PR #164, opened 2026-06-28
#162: prove the extracted RV64IM decoder in-build (raw word → opcode), kernel-soundly

codygunton:
**Closes #162 modulo `native_decide`.** (The single residual — kernel-soundness of the explicit Sail match — is tracked in #174 under #75; see "Native_decide boundary" below.)

## What this proves

ZisK's **real extracted RV64IM byte decoder** (`decode_32_core` / `extract_rv64im_opcode_supported` in `trust/aeneas/ProductionM2.lean`, raw 32-bit word → opcode) now has in-build theorems at three strengths:

1. **Acceptance** — `zisk_decoder_accepts_supported_shape`: every `SupportedDecodeShape` word is accepted (`extract_rv64im_opcode_supported (toU32 raw) = ok true`). *(original)*
2. **Classification (all 63 opcodes)** — `Decode/Classify.lean` per-family `*_classifies` lemmas + headline `zisk_decoder_classifies_supported_shape`: every supported-shape word decodes to a record whose `.opcode` is the **specific `RiscvOpcode`** the RISC-V encoding assigns (pinned to the decoder by `rfl` via `rtypeOpcode`/`itypeOpcode`/…). This is the opcode-classification theorem review finding #2 asked to export.
3. **Explicit Sail-spec match** — `ZiskClassifyMatch.lean` `zisk_decoder_classifies_matches_sail`: for every raw word the **real Sail decoder** decodes to a supported RV64IM `inst`, ZisK classifies it to exactly `inst`'s opcode (`riscvOpcodeOfSailInstr inst`). This is the literal "matching the Sail/RISC-V spec decode" of #162's title.

Completes the chain **bytes → opcode (this PR) → fields (#160)**, all from the real Rust via Aeneas.

## Soundness / native_decide boundary

- **Classification (items 1–2) is fully kernel-sound**: `#print axioms` = `[propext, Classical.choice, Quot.sound]`; no `native_decide`/`bv_decide`/`sorry`. The extraction-closure gate verifies all 741 decode/extraction declarations stay `⊆ {propext, Classical, Quot}`.
[...]

PR #346, opened 2026-08-12
#342: walk the register telescope — cycles excluded on an accepted trace

### LINKED ISSUE #342: Model the register-step descent (bus-102 range slice) — prerequisite for discharging the register fields

codygunton:
Second half of #342. PR #345 modelled `main.pil:333-335`'s bus-102 register-step descent and derived
it from balance. This PR **uses** it: it turns the per-row range fact into a statement about the whole
register partition of an accepted trace.

**Does not close #342** on its own — see "What remains".

## The gap, restated

`a_reg_prev_mem_step` is a free witness column. Two Main rows on one register can point at each
other's read timestamps carrying a fabricated value: a closed 2-cycle touching neither boot nor
reload, with `BalancedChannels` intact. #342's body has the witness.

## What this PR proves

**`registerRead_supplied_by_boundary_or_strictly_later_row`** — on an accepted trace, a Main register
read at `mem_op = 3` is supplied *either* by the `RegisterBoundary`, *or* by a Main row whose own
register access sits at a **strictly later** memory-bus timestamp.

Every premise is discharged from `AcceptedZiskTrace`. Nothing is assumed:

| what | where it comes from |
|---|---|
| the counterpart exists and is register-side | `channels_balanced` → `activeMainMemProviderRowMatchSpec_of_active_main_eval` → `activeMainRegisterProviderRowMatchSpec_of_main_mem_op_three` |
| the supplying row's slot is **active** | its counterpart multiplicity (`≠ 0`) plus Main's own selector booleanity — new `selfMemProvider_registerPre_active_of_mem_op_three` |
| the descent on that row | the bus-102 provider slice (PR #345) |
| the no-wrap bound | the Main table's fixed-column capacity, `mainFixedCapacity = 2^22`, so every read timestamp is below `2^24` |

The last row is worth pausing on. `prev_val_lt_of_registerStepSpec` needs the predecessor timestamp
to be far from the field's wrap point, and the obvious way to get that is to assume the segment is
short. It is not assumed: `Table.index_lt_fixed_capacity` reads the bound off the table carrier
itself (`main_index_lt_mainFixedCapacity`), so it costs no premise and holds for any accepted trace.

**`regSupplies_chain_timestamps_nodup_of_trace`** — consequently the walk visits no read timestamp
twice. The pairing that balance certifies cannot decompose into the boundary path plus a disjoint
cycle among Main rows. That is the configuration #342 exhibits, and `register_walk_no_two_cycle`
refutes its exact two-row shape.

## The walk is slot-indexed on both sides

All three register slots share one MemBus at `mem_op = 3`, so a supply step can **cross** slots: row
`r₁`'s a-side read can be supplied by row `r₂`'s store-side register-pre push. The relation is
therefore `RegSupplies sc sp consumer provider`, indexed by the slot at each end, and the cycle
results exclude mixed-slot cycles too. This replaces the a-only version in this branch's first
commit, which would have left the b/store slots as separate unproved twins.

`RegSlot` collects the three slots' read timestamp, free predecessor column, selector, distance, and
memory-bus message, so the three per-slot lemmas from #345 compose into one
`rangeTable24_spec_distance_of_active` instead of being case-split at every call site.

## The chain result matches what the classification produces

`regSupplies_chain_timestamps_nodup_of_trace` is indexed by `Fin n`, so on its own it speaks only
about executed steps — while `registerRead_counterpart_of_trace` produces providers out of
`witness.allTables`, and a provider may be a **padding row past the executed prefix**. Those two did
not compose, and a chain theorem that cannot consume the classification's output is the same defect
as quantifying over an abstract list.

`IsActiveWitnessMainRow` names the shape the classification actually returns — a Main-component
table of the witness, an in-range index, a slot whose selector is set — the classification returns
it directly, and `regSupplies_chain_timestamps_nodup_of_witnessRows` consumes it. The no-wrap bound
generalizes the same way (`regSlot_timestamp_bound_of_mainTable`), because the fixed-column capacity
argument never cared whether a row was executed or padding. The `Fin n` versions remain as
corollaries.

## Non-vacuity

Every result here is an implication out of `RegSupplies`. If that relation were empty they would all
hold vacuously.

`addX1Row_walk_isChain` exhibits it on the `add x1,x1,x1` witness row — a row of a **checked**
witness, not a hand-built one. Its three slots read at timestamps `1`, `2`, `3`, so the b-slot
supplies the a-slot's read and the store-slot supplies the b-slot's. Those are exactly the
`[0, 0, 0]` bus-102 distances `singleAddWitness`'s provider list records. `addX1Row_walk_timestamps_nodup`
then instantiates the general no-cycle theorem on it.

The chain is a legal strictly-increasing walk within one row, not a cycle — which is the point: the
theorems distinguish the two.

## Trust boundary

**No new trust markers.** No `axiom` / `opaque` / `sorry` / `native_decide` or equivalent. The
non-vacuity witnesses use plain `decide`, which the kernel checks.

V2 **gains** a check rather than losing one: check 19 is the new
`trust/consistency/register_walk_acyclic.lean`, and the Aeneas extraction-pin check moves to 20/20.
Axiom closure of every new theorem is `[propext, Classical.choice, Quot.sound]`; the non-vacuity
chain is `[propext]` alone.

`trust/trusted-base.md`'s register-partition section is updated in the same change: what the descent
now delivers, and — explicitly — what it still does not.

## What remains — this PR does not close #342

1. **Termination at `RegisterBoundary.bootMessage`.** Acyclicity bounds the walk from below, not
   from above. Nothing here forces a chain to *reach* boot, and `main.pil:447`'s reload-timestamp
   check is still unmodelled, so the boundary can self-pair at timestamp `0`.

   Iterating the supply step to prove termination needs one fact this PR does not have: that the
   supplying row's own register access is itself a `mem_op = 3` **pull**, so the counterpart theorem
   applies to it in turn. That needs `a_src_mem + a_src_reg = 1`, because `exists_push_of_pull`
   only fires at multiplicity exactly `-1`.

   Main's constraints give **booleanity only** — `Constraints.lean:252,259` assert
   `a_src_mem * (1 - a_src_mem) = 0` and `a_src_reg * (1 - a_src_reg) = 0`, and there is no
   exclusivity constraint anywhere in the component (checked, not assumed). So `(1, 1)` is
   admissible in the model: the pull would ride at `-2` with `mem_op = 4`. Ruling it out means
   pinning `rom_flags` to a *decoded* instruction, which is the committed-program decode bridge
   (#172, on top of #164's proven decoder) — a different axis from the bus-102 range slice.

   Separately, `main.pil:447` stays unmodelled, so
   `registerBoundary_table_interactionsWith_registerStepRange_nil` is still load-bearing.

   I am flagging both rather than routing them anywhere.

2. **The value telescope.** Ordering is not agreement. Carrying the *value* along the chain, and so
   building #330's 116 `h_a_*_t` / `h_b_*_t` fields instead of assuming them, is #330 Phase 4 and
   sits behind (1).

## Verification

On this branch's head:

- `lake build` — see comment below
- V1 `trust/scripts/check-all.sh` — see comment below
- V2 `trust/scripts/check-all-semantic.sh` (now 20 checks) — see comment below
- no `sorry` anywhere under `ZiskFv/`

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -193,6 +193,7 @@ import ZiskFv.Compliance
 import ZiskFv.Compliance.EnsembleWitnessBuilder
 import ZiskFv.Compliance.RegisterMemBusBalance
 import ZiskFv.Compliance.Instantiation.ConcreteRowReductions
+import ZiskFv.Compliance.RegisterWalk
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Compliance.AddSpinRootSoundness
--- a/ZiskFv/AirsClean/FullEnsemble/Balance/MemBusRowBridges.lean
+++ b/ZiskFv/AirsClean/FullEnsemble/Balance/MemBusRowBridges.lean
@@ -3455,4 +3455,224 @@ theorem selfMemProvider_registerPre_timestamp_of_mem_op_three
       (not_activeMainSelfCMemProviderRowMatchSpec_of_main_mem_op_three
         h_constraints h_mainEval h_main_mem_op)
 
+/-! ### The provider's own slot is active
+
+`selfMemProvider_registerPre_timestamp_of_mem_op_three` identifies *which* message supplied the
+read, but the bus-102 descent is only guaranteed for a slot whose selector is `1` -- an inactive
+slot pulls at multiplicity `0` and carries no range fact. The selector is recoverable: a register-pre
+push rides at multiplicity `<slot>_reg`, `exists_push_of_pull` hands back a counterpart with
+`mult ≠ 0`, and Main's own constraints make that selector boolean. -/
+
+private lemma main_rowInput_a_src_reg_eval
+    (env : Environment FGL) (row : Var ZiskFv.AirsClean.Main.MainRowWithRom FGL) :
+    Expression.eval env row.rom.a_src_reg = (eval env row).rom.a_src_reg := by
+  simp only [ProvableStruct.eval_eq_eval, ProvableStruct.eval,
+    ProvableStruct.fromComponents, ProvableStruct.components,
+    ProvableStruct.toComponents, ProvableStruct.eval.go,
+    ProvableType.eval_field, Expression.eval]
+
+private lemma main_rowInput_b_src_reg_eval
+    (env : Environment FGL) (row : Var ZiskFv.AirsClean.Main.MainRowWithRom FGL) :
+    Expression.eval env row.rom.b_src_reg = (eval env row).rom.b_src_reg := by
+  simp only [ProvableStruct.eval_eq_eval, ProvableStruct.eval,
+    ProvableStruct.fromComponents, ProvableStruct.components,
+    ProvableStruct.toComponents, ProvableStruct.eval.go,
+    ProvableType.eval_field, Expression.eval]
+
+private lemma main_rowInput_store_reg_eval
+    (env : Environment FGL) (row : Var ZiskFv.AirsClean.Main.MainRowWithRom FGL) :
+    Expression.eval env row.rom.store_reg = (eval env row).rom.store_reg := by
+  simp only [ProvableStruct.eval_eq_eval, ProvableStruct.eval,
+    ProvableStruct.fromComponents, ProvableStruct.components,
+    ProvableStruct.toComponents, ProvableStruct.eval.go,
+    ProvableType.eval_field, Expression.eval]
+
+/-- A boolean column is `0` or `1`. -/
+private lemma zero_or_one_of_booleanity {col : FGL} (h : col * (1 - col) = 0) :
+    col = 0 ∨ col = 1 := by
+  obtain ⟨d, hd⟩ := bool_of_booleanity h
+  cases d
+  · exact Or.inl (by simpa [ZiskFv.AirsClean.boolF] using hd)
+  · exact Or.inr (by simpa [ZiskFv.AirsClean.boolF] using hd)
+
+/-- The a-side register-pre push rides at multiplicity `a_src_reg`. -/
+private lemma main_a_reg_pre_eval_mult
+    (env : Environment FGL) (row : Var ZiskFv.AirsClean.Main.MainRowWithRom FGL) :
+    (((MemBusChannel.emitted row.rom.a_src_reg
+      (ZiskFv.AirsClean.Main.aRegPreMessageExpr row)).toRaw).eval env).mult
+      = (eval env row).rom.a_src_reg := by
+  change env row.rom.a_src_reg = (eval env row).rom.a_src_reg
+  exact main_rowInput_a_src_reg_eval env row
+
+/-- The b-side register-pre push rides at multiplicity `b_src_reg`. -/
+private lemma main_b_reg_pre_eval_mult
+    (env : Environment FGL) (row : Var ZiskFv.AirsClean.Main.MainRowWithRom FGL) :
+    (((MemBusChannel.emitted row.rom.b_src_reg
+      (ZiskFv.AirsClean.Main.bRegPreMessageExpr row)).toRaw).eval env).mult
+      = (eval env row).rom.b_src_reg := by
+  change env row.rom.b_src_reg = (eval env row).rom.b_src_reg
+  exact main_rowInput_b_src_reg_eval env row
+
+/-- The store-side register-pre push rides at multiplicity `store_reg`. -/
+private lemma main_c_reg_pre_eval_mult
+    (env : Environment FGL) (row : Var ZiskFv.AirsClean.Main.MainRowWithRom FGL) :
+    (((MemBusChannel.emitted row.rom.store_reg
+      (ZiskFv.AirsClean.Main.cRegPreMessageExpr row)).toRaw).eval env).mult
+      = (eval env row).rom.store_reg := by
+  change env row.rom.store_reg = (eval env row).rom.store_reg
+  exact main_rowInput_store_reg_eval env row
+
+/-- Main's three register selectors are boolean at any row satisfying the component's constraints. -/
+private lemma main_reg_selectors_boolean
+    {length : ℕ} {program : Program length}
+    {table : Table FGL} {row : Array FGL}
+    (h_component :
+      table.component = ZiskFv.AirsClean.Main.componentWithRomMemAndOpBus length program)
+    (h_constraints : table.Constraints)
+    (h_row : row ∈ table.table) :
+    (eval (table.environment row)
+        (ZiskFv.AirsClean.Main.componentWithRomMemAndOpBus length program).rowInputVar).rom.a_src_reg
+        * (1 - (eval (table.environment row)
+          (ZiskFv.AirsClean.Main.componentWithRomMemAndOpBus
+            length program).rowInputVar).rom.a_src_reg) = 0
+      ∧ (eval (table.environment row)
+        (ZiskFv.AirsClean.Main.componentWithRomMemAndOpBus
+          length program).rowInputVar).rom.b_src_reg
+        * (1 - (eval (table.environment row)
+          (ZiskFv.AirsClean.Main.componentWithRomMemAndOpBus
+            length program).rowInputVar).rom.b_src_reg) = 0
+      ∧ (eval (table.environment row)
+        (ZiskFv.AirsClean.Main.componentWithRomMemAndOpBus
+          length program).rowInputVar).rom.store_reg
+        * (1 - (eval (table.environment row)
+          (ZiskFv.AirsClean.Main.componentWithRomMemAndOpBus
+            length program).rowInputVar).rom.store_reg) = 0 := by
+  have h_holds :
+      (ZiskFv.AirsClean.Main.componentWithRomMemAndOpBus length program).operations.ConstraintsHold
+        (table.environment row) := by
+    have := h_constraints row h_row
+    rwa [h_component] at this
+  obtain ⟨-, -, -, -, -, -, -, -, -, -, -, h_a, h_b, h_c⟩ :=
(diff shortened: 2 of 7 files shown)


codygunton:
## Gates on `b0091164`

```
lake build                            Build completed successfully (9161 jobs)
```

The new V2 check reports:

```
::: 19/20 register walk acyclicity (#342) :::
'register_walk_no_two_cycle'                                 [propext, Classical.choice, Quot.sound]
'register_walk_timestamps_nodup'                             [propext, Classical.choice, Quot.sound]
'register_walk_non_vacuous'                                  [propext]
'addX1Row_walk_timestamps_nodup'                             [propext, Classical.choice, Quot.sound]
'registerRead_counterpart_of_trace'                          [propext, Classical.choice, Quot.sound]
'registerRead_supplied_by_boundary_or_strictly_later_row'    [propext, Classical.choice, Quot.sound]
'regSlot_descent_of_trace'                                   [propext, Classical.choice, Quot.sound]
'regSlot_timestamp_bound_of_trace'                           [propext, Classical.choice, Quot.sound]
```

V1's channel-declaration snapshot is unchanged at 71 declarations — this PR adds no channel.

codygunton:
### A lead on the termination blocker, recorded but **not** attempted

Blocker (1) above says `(1, 1)` for `(a_src_mem, a_src_reg)` is admissible in the model, so I routed
it to the committed-program decode bridge. There may be a cheaper route that stays inside balance,
and it is worth writing down before it is forgotten:

If a row sets both flags, its a-side current emission is `emitted (-2) (aMemMessage …)` with
`mem_op = a_src_mem + 3 * a_src_reg = 4`. `BalancedInteractions` in Clean is **message-exact** —
`∀ msg, balanceOf interactions msg = 0`, not a challenge-mixed sum — so that `-2` has to be
cancelled by interactions carrying *that same message*, including `mem_op = 4`. Main's own pushes
are register-pre at `mem_op = 3`, and the memory-side components carry load/store codes. If no
component in the ensemble can emit `mem_op = 4`, balance itself excludes `(1, 1)` and the flag
exclusivity is free.

Two things stand between that sketch and a proof, which is why I have not claimed it:

- Clean's `exists_push_of_pull` fires only at `mult = -1`. This needs a companion at `mult = -2`
  (or a general "a nonzero multiplicity forces a counterpart" form).
- It needs the same silent-on-`mem_op = 4` classification across every component that the bus-102
  work needed on bus 102 — nine nil lemmas' worth, but for a different predicate.

Recording it as a lead, not a result.

codygunton:
Refinement to the lead above, because the first version was too optimistic in one place.

`mem_op = 4` **is** reachable inside Main, so "no component emits it" is wrong as stated:

- a-side: `aMemOp = a_src_mem + 3·a_src_reg` = 4 at `(1, 1)` — the case in question, a **pull** at `-2`;
- b-side: `bMemOp = (b_src_mem + b_src_ind) + 3·b_src_reg` = 4 at `(1, 0, 1)` — also a **pull** at `-2`;
- store side: `cMemOp = 2·(store_mem + store_ind) + 3·store_reg` = 4 at `(1, 1, 0)` — a **push**.

So a positive counterpart does exist in principle. What appears to save the argument is the
timestamp, not the opcode. The three slots sit in fixed residue classes mod 4 —
`1 + 4·main_step`, `2 + 4·main_step`, `3 + 4·main_step` — and `main_step` is the row index, capped
at `2^22`, so no wraparound can move between classes. An a-side pull therefore sits at
`ts ≡ 1 (mod 4)`, while the only `mem_op = 4` push sits at `ts ≡ 3 (mod 4)`. The pushes that *can*
land on `ts ≡ 1 (mod 4)` are the register-pre ones, whose timestamps are free — but those carry
`mem_op = 3`.

That still points at balance rather than at #172, and the residue argument reuses machinery this PR
already has (`main_step` = index, capacity `2^22`). It needs the `mult = -2` counterpart lemma and a
`mem_op = 4` classification, which is comparable in size to the bus-102 work itself. Still a lead,
not a result — but a sharper one.

codygunton:
## Gates on the final head `0fa6918a`

```
lake build                            Build completed successfully (9161 jobs)
```

Check 19 now also covers the witness-rows form:

```
::: 19/20 register walk acyclicity (#342) :::
'register_walk_timestamps_nodup_on_witness_rows'             [propext, Classical.choice, Quot.sound]
'register_walk_no_two_cycle'                                 [propext, Classical.choice, Quot.sound]
'register_walk_timestamps_nodup'                             [propext, Classical.choice, Quot.sound]
'register_walk_non_vacuous'                                  [propext]
'addX1Row_walk_timestamps_nodup'                             [propext, Classical.choice, Quot.sound]
'registerRead_counterpart_of_trace'                          [propext, Classical.choice, Quot.sound]
'registerRead_supplied_by_boundary_or_strictly_later_row'    [propext, Classical.choice, Quot.sound]
'regSlot_descent_of_trace'                                   [propext, Classical.choice, Quot.sound]
'regSlot_timestamp_bound_of_trace'                           [propext, Classical.choice, Quot.sound]
```

Supersedes the gate numbers reported for `b0091164` above; those were taken before the seam commit.

Issue #348, opened 2026-08-12
Model main.pil:447's reload-timestamp range check, so the register walk anchors at bootMessage


codygunton:
Split out of #342, whose register-walk work is now complete and merged.

## The gap

`RegisterBoundary` models two boundary emissions on the register MemBus partition:

- **boot** — `mem.pil:507-508`, per-register call site `main.pil:535-537`: a pull at the literal
  `(timestamp 0, value 0)`;
- **reload** — `main.pil:450`, `reg_pre_load` "Proves the last access.": a push at
  `(reloadTimestamp, reloadValue)`.

`reloadTimestamp` and `reloadValue` are **free** columns. The real circuit constrains the timestamp,
one line above the reload:

```
// main.pil:447
range_check(expression: last_segment_reg_mem_step - last_reg_mem_step[ireg] - 1,
            min: 0, max: MAX_RANGE);
```

with `last_segment_reg_mem_step = main_step_to_special_mem_step((main_segment + 1) * N - 1)`. We do
not model it. `ZiskFv/AirsClean/RegisterBoundary.lean` emits nothing on bus 102, and
`registerBoundary_table_interactionsWith_registerStepRange_nil` currently *proves* that component
silent there — so modelling `447` means revisiting that lemma and the provider case split in
`exists_registerStepRange_provider_of_pull`.

## Why it matters

#342 (merged, PRs #345 / #346 / #347) proved the register walk **terminates**:
`exists_boundarySuppliedSite` says that from any witness site, following supply counterparts reaches
a site whose register read is supplied by `RegisterBoundary`.

That says the walk *reaches* the boundary. It does not say **which** boundary message it lands on.
With `reloadTimestamp` free, the boundary can self-pair — the boot pull at timestamp `0` matched by a
reload push also at timestamp `0` — so nothing yet anchors a register chain at `bootMessage`'s
literal `(ts 0, value 0)`.

Anchoring there is what the **value** telescope needs: ordering is not agreement, and carrying the
register value back to a known `0` at boot is how #330's 116 `h_a_*_t` / `h_b_*_t` fields get
*built* rather than assumed.

## Scope

- Model `main.pil:447` as a bus-102 emission from `RegisterBoundary` (the channel, the 24-bit
  provider slice and `rangeTable24` all already exist from #345).
- Revisit `registerBoundary_table_interactionsWith_registerStepRange_nil` and the provider case
  split, both of which currently rely on `RegisterBoundary` being silent on bus 102.
[...]

PR #347, opened 2026-08-12
#342: source exclusivity from balance, and the register walk terminates at the boundary

codygunton:
Third instalment on #342, after #345 (model the bus-102 descent) and #346 (walk it — cycles
excluded). This one proves **termination**, and along the way removes an obstacle I had wrongly
attributed to a different axis.

## What this PR proves

**`exists_boundarySuppliedSite`** — from any witness site (a Main row with an active register slot),
following supply counterparts reaches a site whose own register read is supplied by
`RegisterBoundary`.

Three steps, each discharged from `AcceptedZiskTrace` with no new premise.

### 1. Operand-source exclusivity, from `BalancedChannels` alone

The walk can only step from a supplying row if that row's own access is a `mem_op = 3` **pull**:
Clean's `exists_push_of_pull` fires only at multiplicity exactly `-1`. Main asserts **booleanity
only** for the source flags (`Main/Constraints.lean:252,259`) and no exclusivity constraint exists in
the component.

`Air.Flat.BalancedInteractions` is **message-exact** — `∀ msg, balanceOf interactions msg = 0`,
summing multiplicities over interactions carrying that exact message array. It is not a
challenge-mixed sum, so a single message can be reasoned about on its own. A row putting two sources
on one slot emits its current access at an opcode nothing else can offset:

| emitter | `mem_op` it can carry |
|---|---|
| Main, three register-pre pushes (the only positive multiplicities) | literal `3` |
| `RegisterBoundary` boot / reload | literal `3` |
| `MemAlignReadByte` | literal `1` |
| `MemAlign`, `Mem` | `wr + 1`, `wr` boolean by their own `Spec` |
| `MemAlignByte` | `1 + is_write`, bounded by its own `Spec` |
| Main, three **current** accesses — all pulls | `0…7` |

So the balance at that message is `m * count` with `count ≥ 1` and `count < GL_prime`, which cannot
vanish (`no_balanced_message_with_constant_nonzero_mult`).

- `mem_op = 4` → `main_not_a_src_mem_and_a_src_reg`
- `mem_op = 7` → `main_not_store_mem_and_store_ind_and_store_reg`
- `mem_op = 5` is reached by **two** emissions at **different** multiplicities (`-3` b-side, `-2`
  store-side), so the count argument does not apply directly. The message **timestamp** separates
  them (`cMem_message_ne_bMem_message`): the slots access at `1, 2, 3 + 4 * main_step`, `main_step`
[...]

PR #349, opened 2026-08-12
#342 follow-up: record the register walk's path, and tighten check 19

codygunton:
Post-merge review follow-ups on #346 and #347. **No proof obligation is discharged here, and none is
weakened.** The changes strengthen one statement, delete surface that #347 superseded, make a trust
gate check what it claims to check, and correct the ledger.

I reviewed both PRs and found no correctness defect. In particular I re-checked and confirm:

* the `RegSlot` abstraction matches the circuit — read timestamps against `Main/Constraints.lean:363,376,389`,
  predecessor columns against `:400,411,422`, `memMult` against the six `MemBusChannel.emit` calls,
  booleanity against `:252,259`, and `distance_a/b/c` are `rfl` against the pre-existing
  `aRegStepDistance` family, so the new names cannot drift from the old lemmas;
* `memBus_mult_eq_of_msg_eq_mem_op_high` case-splits over **all sixteen** components of
  `component_mem_fullRv64im_cases`, not a chosen subset;
* the no-wrap bound really does come from `Table.index_lt_fixed_capacity`, so it costs no premise.

## 1. `exists_boundaryWalk` records the path

This is the one change I would not leave for later.

`exists_boundarySuppliedSite` concluded `BoundarySuppliedSite trace multiplicity as` — a bare
existential over the whole witness that never mentions the `table`/`row`/`s` you started from. So
the theorem stated *"if one active register site exists, then some active register site is
boundary-supplied"*, not *"this site's read traces back to the boundary"*. The proof does follow the
walk; only the statement discarded it. Nothing downstream could then transport a register value,
because the telescope needs the intermediate steps — and `regSupplies_chain_timestamps_nodup_of_witnessRows`
had no way to compose with it.

The fuel induction now returns the walk:

```lean
∃ path : List RegWalkStep, ∃ last : RegWalkStep,
  path.head? = some (eval (table.environment row) …rowInputVar, s)
  ∧ path.getLast? = some last
  ∧ (∀ q ∈ path, IsActiveWitnessMainRow trace q)
  ∧ List.IsChain RegWalkStep.SuppliedBy path
  ∧ BoundarySuppliedAt trace last
```

`site_step` now returns the `RegSupplies` link as well as the timestamp inequality — that is what
makes consecutive steps assemble into an `IsChain` rather than an unrelated sequence of sites.
`boundaryWalk_timestamps_nodup` applies the existing no-cycle theorem to the path, so the walk is
a finite acyclic chain, stated over exactly the rows the classification produces.

The old bare existential survives as `exists_boundarySuppliedSite`, with a docstring saying it is
the weak corollary.

## 2. `RegWalkStep.Supplies` → `RegWalkStep.SuppliedBy`

`def RegWalkStep.Supplies (p q) := RegSupplies p.2 q.2 p.1 q.1` unfolds to "**q** supplies p's
read", so `p.Supplies q` read exactly backwards under dot notation. `p.SuppliedBy q` reads right,
and the chain direction (consumer to provider, forwards in time) is now legible at each use site.
`RegSupplies` keeps its name and gains a line fixing the argument order in prose.

## 3. Two parameters that mean nothing

`ActiveMainRegisterBoundaryProviderRowMatchSpec` `_`-binds its `multiplicity` and `as` arguments.
`BoundarySuppliedSite` copied both into a new definition, and they propagated into
`exists_boundarySuppliedSite` and the trust file. `BoundarySuppliedAt` passes `0 0` and says why in
its docstring.

## 4. Superseded surface deleted

* `registerRead_supplied_by_boundary_or_strictly_later_row` — #346's headline theorem. It takes
  `h_pull : mult = -1` and `h_mem_op = 3` as **hypotheses**, which #347's `site_step` discharges
  from source exclusivity. Nothing called it. Keeping it invites a reader to build on the weaker
  form.
* `traceWalkStep_timestamp_lt_of_supplies`, `not_traceWalkStep_supplies_self` — zero uses.

`not_traceWalkStep_two_cycle` stays: the trust file consumes it.

## 5. V2 check 19 now asserts the axiom closure

`run_lean_no_sorry` fails only on a nonzero exit or ``uses `sorry` ``. The `#print axioms` reports
were printed and never asserted, so check 19 certified *compiles, no sorry* — not the closure the
PR described. `run_lean_no_sorry_std_axioms` requires every report to sit inside
`{propext, Classical.choice, Quot.sound}`, flattening Lean's line-wrapped output first (long
reports wrap over three lines, which a naive line grep gets wrong in both directions).

**Flagging explicitly:** this edits `trust/scripts/`, a protected trust-policy file. It only makes
the gate stricter — it cannot hide anything — but it is your call.

## 6. `main_not_a_src_mem_and_a_src_reg` calls its own helper

It inlined the whole body of `no_balanced_message_with_constant_nonzero_mult`, which its b-side and
store-side twins already call. −35 lines, same proof.

## 7. Ledger

* Removed an orphan sentence: *"…and two things still stand between that and termination"* — #347
  deleted the list and left the lead-in.
* Recorded that the exclusivity count is a **closed-world** argument over the modeled ensemble's
  sixteen components, so it is exactly as strong as that ensemble's completeness with respect to
  the real circuit's memory-bus emitters. That is the standing extraction-fidelity boundary, not a
  new assumption, but this is the first result whose whole content rests on it.
* Restated termination in terms of the path, and pointed the remaining `main.pil:447` gap at #348.

## Verification

On `7141d201`:

* `lake build` — 9163 jobs, green
* `trust/scripts/check-all.sh` — 20/20
* `trust/scripts/check-all-semantic.sh` — 20/20, with check 19 now asserting closures
* no `sorry` under `ZiskFv/`; no new `axiom` / `opaque` / `native_decide`
* axiom closure of every theorem in `trust/consistency/register_walk_acyclic.lean` is
  `[propext, Classical.choice, Quot.sound]`, and `register_walk_non_vacuous` is `[propext]` alone

Details in the comments below.

### DIFF
--- a/ZiskFv/AirsClean/FullEnsemble/Balance/MemBusSourceExclusivity.lean
+++ b/ZiskFv/AirsClean/FullEnsemble/Balance/MemBusSourceExclusivity.lean
@@ -735,39 +735,12 @@ theorem main_not_a_src_mem_and_a_src_reg
   have h_I_mult : I.mult = -2 :=
     main_aMem_mult_eq_neg_two_of_mem_op_four
       (by rw [h_src_mem]; ring) (by rw [h_src_reg]; ring) h_ref_op
-  -- every interaction on that message rides at `-2`
-  have h_all :
-      ∀ i ∈ witness.interactionsWith MemBusChannel.toRaw,
-        i.msg = I.msg → i.mult ≠ 0 → i.mult = -2 := by
-    intro i h_i h_msg _
-    exact memBus_mult_eq_neg_two_of_msg_eq_mem_op_four h_constraints h_specs h_ref_op h_i h_msg
-  have h_balance := memBus_balanced_of_witness witness h_balanced
-  have h_zero := h_balance.2 I.msg
-  rw [balanceOf_eq_of_mult_or_zero h_all] at h_zero
-  set count : ℕ :=
-    (witness.interactionsWith MemBusChannel.toRaw).countP
-      (fun i => i.msg = I.msg && i.mult ≠ 0) with h_count
-  -- the reference interaction itself is counted
-  have h_count_pos : 0 < count := by
-    rw [h_count, List.countP_pos_iff]
-    exact ⟨I, h_I_mem, by simp [h_I_mult]⟩
-  -- and the count is below the characteristic
-  have h_count_lt : count < ringChar FGL ∨ ringChar FGL = 0 := by
-    rcases count_lt_ringChar_of_balancedInteractions (msg := I.msg) h_balance with h | h
-    · exact Or.inl (lt_of_le_of_lt (List.countP_and_left_le _ _ _) h)
-    · exact Or.inr h
-  -- `-2 * count = 0` with `2` invertible forces `count = 0`
-  have h_cast : ((count : ℕ) : FGL) = 0 := by
-    have h2 : (-2 : FGL) ≠ 0 := by decide
-    rcases mul_eq_zero.mp h_zero with h | h
-    · exact absurd h h2
-    · exact h
-  rw [show ringChar FGL = GL_prime from ringChar.eq FGL GL_prime] at h_count_lt
-  have h_count_zero : count = 0 := by
-    rcases h_count_lt with h_lt | h_char
-    · exact natCast_eq_zero_of_lt_prime h_lt h_cast
-    · exact absurd h_char (by decide)
-  omega
+  -- every interaction on that message rides at `-2`, so the balance is `-2 * count` with
+  -- `count ≥ 1`, which cannot vanish
+  refine no_balanced_message_with_constant_nonzero_mult h_balanced (m := -2) (by decide)
+    h_I_mem h_I_mult ?_
+  intro i h_i h_msg _
+  exact memBus_mult_eq_neg_two_of_msg_eq_mem_op_four h_constraints h_specs h_ref_op h_i h_msg
 
 
 /-- **A register-reading row does not also read memory on the a side.** Immediate from
--- a/ZiskFv/Compliance/Instantiation/ConcreteRowReductions.lean
+++ b/ZiskFv/Compliance/Instantiation/ConcreteRowReductions.lean
@@ -2190,7 +2190,10 @@ end RegSlot
 
 /-- `(provider, sp)` supplies `(consumer, sc)`'s register read: the provider slot's register-pre
 push carries the consumer slot's read timestamp. This is exactly what
-`selfMemProvider_registerPre_timestamp_of_mem_op_three` concludes from balance. -/
+`selfMemProvider_registerPre_timestamp_of_mem_op_three` concludes from balance.
+
+Read the argument order as *consumer first*: `RegSupplies sc sp consumer provider` is a fact about
+`consumer`'s read, and `provider` is what answers it. -/
 def RegSupplies (sc sp : RegSlot) (consumer provider : MainRowWithRom FGL) : Prop :=
   sp.prevStep provider = sc.readTimestamp consumer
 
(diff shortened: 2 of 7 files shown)


codygunton:
## Gate evidence on `7141d201`

### `lake build`
```
✔ [9162/9163] Built ZiskFv (3.6s)
Build completed successfully (9163 jobs).
```

### V1 — `trust/scripts/check-all.sh`
```
trust-gate: ALL CHECKS PASSED.
```
(20/20. A fresh worktree needs `git submodule update --init zisk` first, or check 14 fails on a
missing `zisk/core/src/aeneas_extract.rs`.)

### V2 — `trust/scripts/check-all-semantic.sh`
```
::: 19/20 register walk acyclicity + termination (#342) :::
::: 20/20 Aeneas extraction-pin raw axiom closure (V2) :::
trust-gate (V2 semantic): ALL CHECKS PASSED.
```

### Axiom closures under check 19

`trust/consistency/register_walk_acyclic.lean`, now asserted rather than printed:

```
'register_walk_terminates_at_boundary'            [propext, Classical.choice, Quot.sound]
'register_access_is_a_pull'                       [propext, Classical.choice, Quot.sound]
'register_walk_timestamps_nodup_on_witness_rows'  [propext, Classical.choice, Quot.sound]
'register_walk_no_two_cycle'                      [propext, Classical.choice, Quot.sound]
'register_walk_timestamps_nodup'                  [propext, Classical.choice, Quot.sound]
'register_walk_non_vacuous'                       [propext]
'ZiskFv.Compliance.addX1Row_walk_timestamps_nodup'              [propext, Classical.choice, Quot.sound]
'ZiskFv.Compliance.registerRead_counterpart_of_witnessTable'    [propext, Classical.choice, Quot.sound]
'ZiskFv.Compliance.site_step'                                   [propext, Classical.choice, Quot.sound]
'ZiskFv.Compliance.exists_boundaryWalk'                         [propext, Classical.choice, Quot.sound]
'ZiskFv.Compliance.boundaryWalk_timestamps_nodup'               [propext, Classical.choice, Quot.sound]
'ZiskFv.Compliance.regSlot_descent_of_trace'                    [propext, Classical.choice, Quot.sound]
'ZiskFv.Compliance.regSlot_timestamp_bound_of_trace'            [propext, Classical.choice, Quot.sound]
```

`register_walk_terminates_at_boundary` is the path-recording statement now, not the bare
existential, and `site_step` / `exists_boundaryWalk` / `boundaryWalk_timestamps_nodup` are new
entries in the list.

### The new assertion is not vacuous

Negative test of the matcher, including Lean's three-line wrapping:

```
input:
  'A' depends on axioms: [propext,
   Classical.choice,
   Quot.sound]
  'B' depends on axioms: [propext, ZiskFv.someAxiom, Quot.sound]

offenders reported:
  depends on axioms: [propext, ZiskFv.someAxiom, Quot.sound]
```

The wrapped report is accepted; the foreign axiom is caught. The helper also fails if no
`#print axioms` report appears at all, so deleting the reports cannot silence it.

codygunton:
Reviewed as author of #347. **Approved** — all seven fixes are real, and two of them make the result
stronger rather than tidier.

## Verified independently, not taken on report

Detached onto `7141d201` in a separate worktree and ran the full chain myself:

```
lake build                            Build completed successfully (9163 jobs)
```

No `sorry`, and no `axiom` / `opaque` / `native_decide` / `@[extern]` added anywhere in the
`ZiskFv/` diff. Axiom closure of the new surface (`exists_boundaryWalk`,
`boundaryWalk_timestamps_nodup`, `site_step`, `exists_boundarySuppliedSite`) is the three standard
axioms. None of the six deleted names is referenced anywhere in `ZiskFv/` or `trust/`.

## On the two changes worth arguing about

**Fix 3 — dropping `multiplicity` / `as` from the boundary predicate.** I checked the claim rather
than accepting it: `ActiveMainRegisterBoundaryProviderRowMatchSpec` really does bind both as
`_multiplicity _as` and never mentions them in its body
(`MemBusRowBridges.lean:3107-3121`). So passing `0 0` loses nothing, and my version was carrying two
parameters that meant nothing. Correct call.

**Fix 5 — editing `trust/scripts/`, a protected trust-policy file.** The rule this could violate is
"do not edit trust policy to silence a gate". This does the opposite, and I tested that it bites
rather than assuming it:

- pointed `run_lean_no_sorry_std_axioms` at `root_soundness`, whose closure includes the Sail
  axioms → **exit 1**, `Axiom closure outside {propext, Classical.choice, Quot.sound}`;
- pointed it at a file with no `#print axioms` at all → **exit 1**, `No '#print axioms' report
  found`. So the check cannot be passed by deleting the reports.

Check count is unchanged at 20; only check 19 is upgraded from printing to asserting. Approved.

## The substantive win

Fix 1 is not a cleanup. `exists_boundaryWalk` returns the path with `head? = some` **the site you
started from**, every element an `IsActiveWitnessMainRow`, chained by `SuppliedBy`, ending at
`BoundarySuppliedAt`. My `exists_boundarySuppliedSite` only said *some* site is boundary-supplied —
which, as the PR body says, does not connect the boundary to the read you began at, and the value
telescope needs the intermediate steps. The bare existential survives as a corollary, correctly
labelled as the weak one.

Fix 2 is also more than naming: under dot notation `p.Supplies q` read as "p supplies q" while the
definition meant the reverse. `SuppliedBy` matches the walk's actual direction, consumer to provider,
forwards in time.

## One thing I'd missed, now in the ledger

The closed-world caveat added to `trust/trusted-base.md`: the count argument case-splits over the
sixteen components of `component_mem_fullRv64im_cases`, so it is exactly as strong as the ensemble's
completeness with respect to the real circuit's memory-bus emitters. That is the standing
extraction-fidelity boundary, but this is the first result whose entire content rests on it, and it
deserved saying. I had not written that down.

PR #350, opened 2026-08-13
#348: the register walk lands on the reload, so the boundary cannot self-pair

codygunton:
Resolves the outcome #348 was opened for, and **corrects that issue's premise**, which was wrong.

## What #348 wanted

`exists_boundaryWalk` (#347) says a register chain ends at the `RegisterBoundary`. That component
emits two messages, so "ends at the boundary" left a question open: **which** one? If the reload
could answer the boot pull, the boundary would self-pair and nothing would anchor a chain at
`bootMessage`'s literal `(ts 0, value 0)`.

## It is the reload, and no new constraint was needed

`RegisterBoundary` emits its boot pull at multiplicity `-1` and its reload push at `+1`
(`RegisterBoundary.lean:126-128`). `ActiveMainRegisterBoundaryProviderRowMatchSpec` — the shape
`exists_push_of_pull` hands back — carries `mult ≠ -1`. **The boot pull is excluded by
construction.**

- `boundarySuppliedAt_reload_timestamp` — the walk's landing point is the reload, and balance
  equates the reload's whole message with the read's, so `reloadTimestamp` equals a real Main read
  timestamp.
- `readTimestamp_val_pos` — every Main read timestamp is `k + 4 * index` with `k ∈ {1, 2, 3}` and
  `index < mainFixedCapacity`, hence **positive**.
- `boundary_reload_ne_boot` — therefore the reload can never answer the boot pull's `timestamp = 0`
  message. **The boundary cannot self-pair** whenever any register read occurs.

Axiom closure of all three: `[propext, Classical.choice, Quot.sound]`. No `sorry`, no new trust
marker. V2 check 19 certifies the result.

## The correction

#348 says `main.pil:447` is what rules the self-pairing out. **It does not.**

`main_step_to_special_mem_step(step) = RESERVED_MEM_STEPS + 4 * step + 4 - 1 = 1 + 4 * step + 3`
(`mem.pil:520-521`, `RESERVED_MEM_STEPS = 1`, `MAX_MEM_STEPS_PER_MAIN_STEP = 4`). So
`last_segment_reg_mem_step = 4 * N` for a segment of `N` rows, and with
`N = mainFixedCapacity = 2^22` that is `2^24`. At `reloadTimestamp = 0` the expression `447`
range-checks is `2^24 - 0 - 1 = 2^24 - 1`, which is **inside** `rangeTable24`. `447` admits `0`.

The exclusion comes from the multiplicity ledger instead — the same place operand-source exclusivity
came from in #347. This is the third time in this stream that a gap I attributed to a missing
constraint turned out to follow from balance; the ledger now records it.

## What I did *not* build, and why

`447` is still genuinely unmodelled, and it is still an extraction-fidelity gap. What it would buy
is an **upper** bound, `reloadTimestamp ≤ 2^24 - 1`.

Nothing in the development consumes that bound today. The backward walk to `bootMessage` — the next
step, and #330 Phase 4's real dependency — will not consume it either: the reload is a *push*, so it
is never the target of a backward step, which goes push-to-pull.

Modelling it is not free. `RegisterBoundary` would become a bus-102 consumer, which deletes
`registerBoundary_table_interactionsWith_registerStepRange_nil` (currently *proved*), changes the
provider case split in `exists_registerStepRange_provider_of_pull`, and adds one provider row **per
register** — 31 in the current witnesses — to all seven accepted-trace witnesses and their balance
proofs.

I did not do that speculatively for a fact with no consumer. It is recorded in
`trust/trusted-base.md` with the cost, so it can be picked up deliberately. **Whether #348 closes on
this is your call**: the outcome it wanted is delivered, its premise is corrected, and the literal
modelling in its title is not built.

## Verification

Gates on this head — see the comment below.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -197,6 +197,7 @@ import ZiskFv.Compliance.Instantiation.ConcreteRowReductions
 import ZiskFv.Compliance.RegisterWalk
 import ZiskFv.Compliance.MemBusSlotSeparation
 import ZiskFv.Compliance.RegisterBoundaryAnchor
+import ZiskFv.Compliance.RegisterBoundaryAnchorNonvacuity
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Compliance.AddSpinRootSoundness
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -196,6 +196,7 @@ import ZiskFv.AirsClean.FullEnsemble.Balance.MemBusSourceExclusivity
 import ZiskFv.Compliance.Instantiation.ConcreteRowReductions
 import ZiskFv.Compliance.RegisterWalk
 import ZiskFv.Compliance.MemBusSlotSeparation
+import ZiskFv.Compliance.RegisterBoundaryAnchor
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Compliance.AddSpinRootSoundness
--- a/ZiskFv/Compliance/Instantiation/ConcreteRowReductions.lean
+++ b/ZiskFv/Compliance/Instantiation/ConcreteRowReductions.lean
@@ -2168,6 +2168,28 @@ def memMult : RegSlot → Var MainRowWithRom FGL → Expression FGL
   | .b, row => -(row.rom.b_src_mem + row.rom.b_src_ind + row.rom.b_src_reg)
   | .c, row => -(row.rom.store_mem + row.rom.store_ind + row.rom.store_reg)
 
+/-- The whole value-level read message of a slot. Keeping it as one object means a consumer that
+needs the *values* — the register telescope — does not have to re-derive them field by field. -/
+def readMessage : RegSlot → MainRowWithRom FGL →
+    ZiskFv.Channels.MemoryBus.MemBusMessage FGL
+  | .a, row => ZiskFv.AirsClean.Main.aMemMessage row
+  | .b, row => ZiskFv.AirsClean.Main.bMemMessage row
+  | .c, row => ZiskFv.AirsClean.Main.cMemMessage row
+
+/-- Evaluating a slot's read message gives that slot's value-level read message. -/
+theorem eval_memMessageExpr (s : RegSlot) (env : Environment FGL)
+    (row : Var MainRowWithRom FGL) :
+    eval env (s.memMessageExpr row) = s.readMessage (eval env row) := by
+  cases s
+  · rw [memMessageExpr, readMessage, ZiskFv.AirsClean.Main.eval_aMemMessageExpr]
+  · rw [memMessageExpr, readMessage, ZiskFv.AirsClean.Main.eval_bMemMessageExpr]
+  · rw [memMessageExpr, readMessage, ZiskFv.AirsClean.Main.eval_cMemMessageExpr]
+
+/-- The read message's own `timestamp` field is the slot's read timestamp. -/
+@[simp] theorem readMessage_timestamp (s : RegSlot) (row : MainRowWithRom FGL) :
+    (s.readMessage row).timestamp = s.readTimestamp row := by
+  cases s <;> rfl
+
 /-- The read message's timestamp is the slot's read timestamp. -/
 theorem eval_memMessageExpr_timestamp (s : RegSlot) (env : Environment FGL)
     (row : Var MainRowWithRom FGL) :
--- a/ZiskFv/Compliance/RegisterBoundaryAnchor.lean
+++ b/ZiskFv/Compliance/RegisterBoundaryAnchor.lean
@@ -101,7 +101,6 @@ theorem boundarySuppliedAt_reload_message
     rw [h_p1]
     exact h_full
 
-
 /-- The reload's timestamp is the read's timestamp — the projection the walk uses. -/
 theorem boundarySuppliedAt_reload_timestamp
     {n : Nat} (trace : AcceptedZiskTrace n) {p : RegWalkStep}
--- a/ZiskFv/Compliance/RegisterBoundaryAnchor.lean
+++ b/ZiskFv/Compliance/RegisterBoundaryAnchor.lean
@@ -1,0 +1,136 @@
+import ZiskFv.Compliance.MemBusSlotSeparation
+
+/-!
+# The register walk lands on the reload, and the reload timestamp is a real read
+
+`exists_boundaryWalk` says a register chain ends at the `RegisterBoundary`. That component emits two
+messages, so "ends at the boundary" leaves a question: **which** one?
+
+It is the **reload**, and nothing new is needed to see it. `RegisterBoundary` emits its boot pull at
+multiplicity `-1` and its reload push at `+1` (`RegisterBoundary.lean:126-128`), while
+`ActiveMainRegisterBoundaryProviderRowMatchSpec` — the shape `exists_push_of_pull` hands back —
+carries `mult ≠ -1`. The boot pull is excluded by construction.
+
+The consequence is worth stating on its own: the reload's free `reloadTimestamp` column **equals a
+real Main read timestamp**, which is `k + 4 * main_step` for `k ∈ {1, 2, 3}` and therefore never `0`.
+So the boundary cannot self-pair — its boot pull cannot be answered by its own reload push —
+whenever any register read occurs.
+
+## On #348
+
+That issue was opened on the belief that `main.pil:447` is what rules the self-pairing out. It is
+not. `447` range-checks `last_segment_reg_mem_step - last_reg_mem_step - 1`, and
+`main_step_to_special_mem_step(step) = 1 + 4 * step + 3` makes `last_segment_reg_mem_step = 4 * N`
+for a segment of `N` rows. With `N = mainFixedCapacity = 2^22` that is `2^24`, so at
+`reloadTimestamp = 0` the checked expression is `2^24 - 1`, which is *inside* `rangeTable24`. `447`
+admits `0`.
+
+The exclusion comes from the multiplicity ledger instead — the same place operand-source exclusivity
+came from. `447` is still a real unmodelled constraint, and it still buys something (an upper bound
+on the reload timestamp, which the backward walk to `bootMessage` will want), but it is not the
+anchor.
+-/
+
+namespace ZiskFv.Compliance
+
+open Air.Flat
+open ZiskFv.AirsClean.FullEnsemble
+open ZiskFv.AirsClean.ZiskInstructionRom (Program)
+open ZiskFv.AirsClean.Main (componentWithRomMemAndOpBus)
+open ZiskFv.Channels.MemoryBus (MemBusChannel)
+open ZiskFv.Compliance.Instantiation (RegSlot RegWalkStep)
+
+/-- The boot pull rides at `-1`. This is the whole reason it cannot be a counterpart. -/
+theorem registerBoundary_boot_eval_mult (env : Environment FGL)
+    (row : Var ZiskFv.AirsClean.RegisterBoundary.RegisterBoundaryRow FGL) :
+    (((MemBusChannel.emitted (-1)
+      (ZiskFv.AirsClean.RegisterBoundary.bootMessageExpr row)).toRaw).eval env).mult = -1 := by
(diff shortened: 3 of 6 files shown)


codygunton:
## Gates on `f28751c5`

```
lake build                            Build completed successfully (9164 jobs)
```

No `sorry` under `ZiskFv/`, no new trust marker. Check 19 — which now *asserts* the closure rather
than printing it (#349) — covers the new result:

```
::: 19/20 register walk acyclicity + termination (#342) :::
'register_walk_boundary_is_reload'   [propext, Classical.choice, Quot.sound]
'register_walk_terminates_at_boundary'  [propext, Classical.choice, Quot.sound]
...
```

codygunton:
## Review of `f28751c5`

Gates re-verified independently in `.worktrees/clean-migration-330` (branch `330-phase4-value-telescope`,
which carries this file content unchanged):

```
lake build ZiskFv.Compliance.RegisterBoundaryAnchor   Build completed successfully (8330 jobs)
lake env lean trust/consistency/register_walk_acyclic.lean
  'register_walk_boundary_is_reload'  [propext, Classical.choice, Quot.sound]
```

**Verdict: the Lean is correct and the theorems say what they claim. Three prose/ledger claims
overreach, and one cheap result is left on the floor.**

### 1. The self-pairing claim is falsified by our own witness

`boundary_reload_ne_boot` proves: for the boundary row that answers *this* read,
`reloadTimestamp ≠ 0`. The docstring, the PR body, and `trust/trusted-base.md` all say instead that
the boundary cannot self-pair "**whenever any register read occurs**". That is a global claim, and it
is false.

`RegisterMemBusBalance.lean:298` is
`boundaryRowIdle r = { reg := r, reloadTimestamp := 0, reloadValue_0 := 0, reloadValue_1 := 0 }` — a
row that self-pairs by construction. `AddFaithfulPaddedWitness.lean:122` puts 30 of them in the
accepted-trace witness, alongside a register read on x1. So a register read occurs and 30 boundary
rows self-pair in the same trace.

The true statement is per-register: **for any register that is read**, its boundary row's reload
cannot answer its own boot pull. Registers never read still self-pair, and nothing depends on them.
Fix the wording in all three places.

### 2. The `447` correction drops `main_segment`

`main.pil:439` is `main_step_to_special_mem_step((main_segment + 1) * N - 1)`, so
`last_segment_reg_mem_step = 4 * (main_segment + 1) * N`, not `4 * N`. `main_segment` is a free
airval, range-checked to `0 .. 2^MAIN_STEP_BITS/N - 1` (`main.pil:73,541`).

At `main_segment = 1` and `N = 2^22`, the expression `447` checks at `last_reg_mem_step = 0` is
`2^25 - 1`, which is **outside** `MAX_RANGE = (1 << 24) - 1` (`main.pil:11`). So `447` **does**
exclude `0` in every segment after the first. The correction holds only at `main_segment = 0`.

The Lean model is single-segment, so the conclusion survives. But the ledger states a PIL-level fact
unconditionally, and a future reader will take it at face value. Qualify it.

### 3. Zero margin at the model's own capacity

At `main_segment = 0` and `N = mainFixedCapacity = 2^22`, the checked expression is exactly
`2^24 - 1 = MAX_RANGE`. `447` admits `0` only because the bound is inclusive — there is no slack.
"Which is *inside* `rangeTable24`" reads as if there is room. One clause fixes it. (The PIL's own
default `N = 2**21` gives `2^23 - 1`, which is genuinely inside; the two differ.)

### 4. The value equality is proved and then discarded

`boundarySuppliedAt_reload_timestamp` builds
`h_full : eval … reloadMessageExpr = eval … memMessageExpr` — **full** `MemBusMessage` equality —
then takes only `.timestamp`. `reloadValue_0` and `reloadValue_1` equal the read's values, one
`congrArg` away.

That is exactly what #330 Phase 4's value telescope needs, and #348 names it as the reason the anchor
matters ("ordering is not agreement"). Expose it here rather than re-deriving it later.

### Minor

- `boundary_reload_ne_boot` destructures `h` and then rebuilds the anonymous constructor to feed
  `boundarySuppliedAt_reload_timestamp`. Destructure a copy and pass `h`.
- Stray double blank line before `readTimestamp_val_pos`.
- `BoundarySuppliedAt` gets no concrete non-vacuity witness, unlike `SuppliedBy`
  (`register_walk_non_vacuous`). It is inhabited in practice — `addFaithfulBoundaryRowX1` is the x1
  reload at a real read timestamp — so it is a cheap addition that matches the standard the file
  already sets.

### On closing #348

The scope question was handled correctly: `main.pil:447` is not modelled, the PR says so plainly, and
the decision was routed to the owner instead of the issue being closed. The outcome #348 wanted is
delivered; its literal title is not built. #348 should stay open with its title and scope rewritten
to the remaining upper-bound work, with a pointer to this PR for the anchor. Closing it would let the
title's deliverable disappear.

codygunton:
Thanks — findings 1 and 2 are errors in what I wrote, not wording quibbles. All four addressed in
`e22e8f12`.

### 1. Self-pairing claim — corrected, it was false as stated

You're right and the counterexample is in our own tree. `boundaryRowIdle`
(`RegisterMemBusBalance.lean:298`) sets `reloadTimestamp := 0` and self-pairs by construction, and
`AddFaithfulPaddedWitness` puts thirty of them in an accepted witness beside a real read on `x1`. So
"cannot self-pair whenever any register read occurs" is refuted by a witness we ship.

The theorem was always the per-register statement — `boundary_reload_ne_boot` quantifies over the
boundary row that answers *this* read. Only the prose overreached. Fixed in the module docstring and
`trust/trusted-base.md`, both of which now say the claim is per register and that rows for
never-read registers do self-pair and carry nothing.

### 2. `main_segment` — dropped, and it changes the conclusion's scope

`main.pil:439` is `main_step_to_special_mem_step((main_segment + 1) * N - 1)`, so
`last_segment_reg_mem_step = 4 * (main_segment + 1) * N`. I wrote `4 * N`.

At `main_segment ≥ 1` the expression `447` checks at `reloadTimestamp = 0` is at least `2^25 - 1`,
outside `MAX_RANGE` — so `447` **does** exclude `0` in every segment after the first. The Lean model
is single-segment so the conclusion holds where it is used, but I stated a PIL-level fact
unconditionally. Both the module docstring and the ledger now carry the segment qualification.

### 3. Zero margin — said plainly now

At `main_segment = 0`, `N = 2^22`, the checked expression is `2^24 - 1`, which *equals* `MAX_RANGE`.
Both places now say "equal to `MAX_RANGE` exactly, admitted with zero margin" rather than "inside".

### 4. The value equality — exposed, and it was the right call

Restructured rather than patched. The primitive is now `boundarySuppliedAt_reload_message`, which
returns the whole `MemBusMessage` equality; `boundarySuppliedAt_reload_timestamp` and the new
**`boundarySuppliedAt_reload_values`** are projections off it. That needed `RegSlot.readMessage` and
`readMessage_timestamp` so the read side is one object rather than field-by-field.

V2 check 19 now also certifies `register_walk_boundary_carries_values`. This is the agreement half —
ordering is not agreement — so #330 Phase 4 consumes it directly instead of re-deriving it.

### Minor

All three done: `boundary_reload_ne_boot` no longer destructures and rebuilds the anonymous
constructor, the stray blank line is gone, and the value-carrying result is certified in the trust
gate.

On the non-vacuity witness for `BoundarySuppliedAt`: I did not add one. Building it needs a concrete
`AcceptedZiskTrace` whose walk actually reaches the boundary, which is more than the row-level
`decide` that `register_walk_non_vacuous` uses. Flagging rather than silently skipping — say the
word and I'll do it.

### On closing #348

Agreed, and I have not touched it. It stays open; the retitle-and-rescope is the owner's call, not
mine to make while the PR is unmerged.

Gates on the new head follow in the next comment.

codygunton:
## Gates on `e22e8f12` (post-review head)

```
lake build                            Build completed successfully (9164 jobs)
```

No `sorry` under `ZiskFv/`, no new trust marker. Check 19 — which asserts the closure rather than
printing it — now covers the value equality as well:

```
::: 19/20 register walk acyclicity + termination (#342) :::
'register_walk_boundary_carries_values'  [propext, Classical.choice, Quot.sound]
'register_walk_boundary_is_reload'       [propext, Classical.choice, Quot.sound]
'register_walk_terminates_at_boundary'   [propext, Classical.choice, Quot.sound]
...
```

Supersedes the numbers reported for `f28751c5`; those predate the review fixes.

codygunton:
## Re-review of `e22e8f12`

Gates re-verified independently in `.worktrees/clean-migration-330`, checked out at the PR head:

```
lake build ZiskFv.Compliance.RegisterBoundaryAnchor   Build completed successfully (8330 jobs)
lake env lean trust/consistency/register_walk_acyclic.lean
  'register_walk_boundary_carries_values'  [propext, Classical.choice, Quot.sound]
  'register_walk_boundary_is_reload'       [propext, Classical.choice, Quot.sound]
```

**All four findings are addressed. Two residual nits, neither blocking.**

### What landed

1. **Self-pairing scope** — fixed in the docstring and the ledger, and the fix goes further than the
   review asked: both now name `boundaryRowIdle` and the thirty idle rows in
   `AddFaithfulPaddedWitness` as the concrete counterexample. Citing the witness that falsifies the
   old wording is the right way to close a claim-scope error.
2. **`main_segment`** — corrected to `4 * (main_segment + 1) * N` with the `main_segment ≥ 1`
   exclusion stated in both places.
3. **Zero margin** — stated as "equal to `MAX_RANGE` exactly … with zero margin".
4. **Value equality** — the restructure is the right shape. `boundarySuppliedAt_reload_message` is
   now the primitive carrying full `MemBusMessage` equality, with `boundarySuppliedAt_reload_timestamp`
   and `boundarySuppliedAt_reload_values` as projections off it. `readMessage` /
   `readMessage_timestamp` (proved `cases s <;> rfl`) keep `readTimestamp` definitionally tied to the
   message field, so the two cannot drift.

The `obtain` round-trip in `boundary_reload_ne_boot` is also gone.

### Residual 1 — `register_walk_boundary_carries_values` is not non-vacuity

Its docstring opens "Non-vacuity for `BoundarySuppliedAt`'s payload". It is not. The theorem is
conditional on `BoundarySuppliedAt trace p` and exhibits no inhabitant. `register_walk_non_vacuous`,
seventy lines above in the same file, exhibits a concrete chain — that is what the word means here,
and reusing it for a conditional projection blurs a distinction this file deliberately maintains.

Call it what it is: the reload carries the read's values. It is a strengthening of the payload, not
evidence that the payload is reachable.

### Residual 2 — `BoundarySuppliedAt` still has no concrete inhabitant

Unchanged from the first round. `addFaithfulBoundaryRowX1` is the x1 reload at a real read timestamp
and `addFaithfulReloadMessage_eq` already proves `reloadMessage addFaithfulBoundaryRowX1 =
cMemMessage addFaithfulRow1` by `rfl`, so the witness is essentially sitting there. Cheap, and it
would make Residual 1's docstring true.

### Nit

Double blank line at `RegisterBoundaryAnchor.lean:103-104`.

### On the ledger heading

"**#348's premise was wrong, and this corrects the record.**" is still flat, though the sentences
after it now carry the qualification. "Wrong for the first segment, which is the case the model is
in" would match the paragraph it introduces.

codygunton:
## Review residuals applied — gates on `35e4079f`

```
lake build                            Build completed successfully (9165 jobs)
```

Check 19 now also asserts the landing point is reached:

```
::: 19/20 register walk acyclicity + termination (#342) :::
'register_walk_boundary_reached'        [propext, Classical.choice, Quot.sound]
'register_walk_boundary_carries_values' [propext, Classical.choice, Quot.sound]
'register_walk_boundary_is_reload'      [propext, Classical.choice, Quot.sound]
...
```

All three review residuals are closed:

1. **`BoundarySuppliedAt` now has a concrete inhabitant.**
   `exists_boundarySuppliedAt_addFaithful` (new file
   `ZiskFv/Compliance/RegisterBoundaryAnchorNonvacuity.lean`) discharges it on
   `addFaithfulAcceptedTrace` by handing row `0`'s `a` slot to
   `exists_boundarySuppliedSite`. The three anchor results are no longer conditional on a
   hypothesis nothing was shown to satisfy.
2. **`register_walk_boundary_carries_values` no longer calls itself non-vacuity.** Its docstring
   states what it proves and points at the witness, leaving `register_walk_non_vacuous`'s meaning of
   the word intact.
3. Stray blank line removed.

The ledger heading now carries the first-segment qualification up front rather than after the fact,
and records the non-vacuity discharge.

No new trust marker, no `sorry`.

Issue #343, opened 2026-08-11
Chain SailTrace so the PC arm is genuinely discharged (protected interface — needs sign-off)


codygunton:
Sub-issue of #330. This is the **only** change that would let #330's PC arm be called discharged rather than restructured.

## Why the restructure is not enough

`SegmentPcChain.retire` exists only because `SailTrace` is a bare `Fin n → SailState` (`ZiskFv/Compliance/SailTrace.lean`) with **no chaining** — nothing says the Sail states form an execution. Because it is an assumption rather than a definition, `sailRetireChain_of_inputsAgree` can derive it back from the old per-row bundle, which is exactly why the PC-arm work is inter-derivable with what it replaced.

## What "done" means

`SailTrace` is defined as the sequence Sail's own semantics generates from an initial state, so that `retire` is **definitional** and `boot` is the only PC premise left on `root_soundness`.

The groundwork is already proved:

- `Pilot.execRowOf trace i`'s producer entry is *definitionally* `mainOfTable.pc (i + 1)`;
- `bus_effect` writes `Register.nextPC := BitVec.ofNat 64 (execution_bus[1]!.pc).val`;
- `mainOfTable_pc_succ_eq_nextPcMux` and `execRowOf_producer_pc_eq_nextPcMux` close the loop.

So `retire j` already says exactly "Sail's PC at `j + 1` is what the circuit wrote into Sail's `nextPC` at step `j`". What is missing is Sail's **retire** step being part of the trace's definition rather than assumed.

## Cost, and why it needs sign-off first

`SailTrace` is a protected interface. It is referenced by all 63 `StepSound` arms, every `Inputs_<op>` structure, and all 7 accepted-trace witnesses. Per `AGENTS.md` this needs explicit owner permission before starting.

`sailInstructionOf` (63 arms) already exists in `SailRetireChain.lean` and names the instruction each `ZiskStep` executes, which is the piece a chained definition needs.

## Guardrail

Moving `retire` into `SailTrace`'s definition is only a genuine discharge if the resulting `SailTrace` is *harder to inhabit* in the right way — i.e. constructing one requires actually running Sail. If it merely repackages the same assumption as a structure field, that is laundering and must be reported as such, exactly as the PC-arm restructure was.

PR #351, opened 2026-08-13
#330 Phase 4 (ZisK half): the register value telescope reaches bootMessage

codygunton:
#330 Phase 4, ZisK half. Stacked on #350 (`348-boundary-anchor`).

Ordering is not agreement. #342 proved the register walk is acyclic and terminates; this carries the
**value** along it, back to a register that still holds `0`.

## The direction the plan never fixed

The forward walk of #342/#347 ends at the `RegisterBoundary` **reload**. The value telescope has to
run the other way — from a row's register-pre push, find the *pull* it answers — and that is a
different walk on a **decreasing** measure, not a reuse of `exists_boundaryWalk_of_fuel`. The plan
now says so.

## What this proves

**`exists_bootAnchored`** — from any active register slot of any witness Main row, following
register-pre pushes back reaches a slot whose predecessor access is the boot pull.
**`bootAnchored_values_zero`** — that slot's operand values are `0` and its predecessor timestamp is
`0`.

The pieces:

- **`exists_pull_of_push_memBus`** — Clean ships `exists_push_of_pull` and no converse. `balanceOf`
  is sign-symmetric, so the converse is the same count argument at `m = 1`, and
  `no_balanced_message_with_constant_nonzero_mult` already had the arithmetic.
- **`memBus_mem_op_three_counterpart`** — the sixteen-component split. A `mem_op = 3` counterpart at
  multiplicity outside `{0, 1}` is a Main current access or the boot pull. Ten components emit
  nothing on this bus; the Mem/MemAlign family carries `mem_op ∈ {1, 2}`; the reload rides at `+1`;
  and Main's three register-pre pushes ride at their own selector.
- **`main_regPre_mult_zero_or_one`** — that last exclusion. **`mult ≠ 0` is load-bearing**: a
  register-pre push on an *inactive* slot is a phantom emission at multiplicity `0` carrying a real
  message, and idle slots are everywhere in the witnesses. It does here what `mult ≠ -1` did for the
  boot pull going forwards.
- **`regSlot_selector_of_mem_op_three`** — a `mem_op = 3` read has its own register selector set
  (`a_src_mem + 3 · a_src_reg = 3` with both boolean forces `a_src_reg = 1`), which is what lets the
  walk take *that* row's register-pre push in turn.
- **`backward_step_lt`** — the strict decrease. Worth saying why it could not come earlier: the
  predecessor column is a **free witness column**, so nothing bounds it *a priori*. The bound
  arrives *with* the step — once the answering read is identified, its timestamp is `k + 4 · index`
  and so below `2 ^ 40`, and the bus-102 descent then puts it strictly below this row's access.

Axiom closure of everything new: `[propext, Classical.choice, Quot.sound]`. No `sorry`, no new trust
marker, no new premise.

## How the chain carries writes

Worth recording, because it looks wrong at first. The a- and b-side register-pre messages carry the
row's **own** operand values, so along an a-side chain the value is *constant* — correct, reads do
not write. The store side is where it changes: `cRegPreMessage` carries `store_reg_prev_value` while
`cMemMessage` carries the new store value. So a write is exactly the link where the chain's value
steps, and the per-step message equality already threads it. Nothing extra was needed.

## What this does **not** do — the honest split

The 116 fields are cross-machine:

```
h_a_lo_t : m.a_0 r_main = lane_lo ((sail_to_rv64 state).xreg (regidx_to_fin r1))
```

This PR delivers the **ZisK half**: the `a_0` column traces back through message equalities to a
register that still holds `0`, with writes threaded along the way. It says nothing about Sail's
`xreg`.

The remaining half is the cross-machine induction the plan describes: widen Phase 7's
`stepSound_of_programDecodes` invariant from `PcAgree j` to `PcAgree j ∧ RegAgree j`, with the step
being writeback preservation off `StepSound j`'s post-state, and a `regs`-projection lemma in the
shape of `nextPC_of_busEffect_ok`. That is not built here.

**The bargain, stated in advance as the plan requires.** `RegAgree 0` becomes a new named boot
premise, in the same class as `SegmentPcChain.boot` and `BootSegmentMemorySeed`. Phase 4 is a real
reduction **only if** the 116 per-row fields go away and that one boot equation replaces them. I
have not reached the point where that exchange is decidable, so nothing in this PR claims it — and
`root_soundness`'s statement is untouched.

## Verification

Gates in the comment below.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -198,6 +198,7 @@ import ZiskFv.Compliance.RegisterWalk
 import ZiskFv.Compliance.MemBusSlotSeparation
 import ZiskFv.Compliance.RegisterBoundaryAnchor
 import ZiskFv.Compliance.RegisterBoundaryAnchorNonvacuity
+import ZiskFv.Compliance.RegisterValueTelescope
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Compliance.AddSpinRootSoundness
--- a/ZiskFv/AirsClean/FullEnsemble/Balance/MemBusSourceExclusivity.lean
+++ b/ZiskFv/AirsClean/FullEnsemble/Balance/MemBusSourceExclusivity.lean
@@ -57,7 +57,7 @@ theorem zero_or_one_of_bool {col : FGL} (h : col * (1 - col) = 0) : col = 0 ∨ 
     linear_combination this
 
 /-- A field element whose `val` is below `2` is `0` or `1`. -/
-private lemma zero_or_one_of_val_lt_two {x : FGL} (h : x.val < 2) : x = 0 ∨ x = 1 := by
+theorem zero_or_one_of_val_lt_two {x : FGL} (h : x.val < 2) : x = 0 ∨ x = 1 := by
   interval_cases h_val : x.val
   · exact Or.inl (by apply Fin.ext; simp [h_val])
   · exact Or.inr (by apply Fin.ext; simp [h_val])
--- a/ZiskFv/Compliance/Instantiation/ConcreteRowReductions.lean
+++ b/ZiskFv/Compliance/Instantiation/ConcreteRowReductions.lean
@@ -2200,6 +2200,28 @@ theorem eval_regPreMessageExpr_mem_op (s : RegSlot) (env : Environment FGL)
   · rw [regPreMessageExpr, ZiskFv.AirsClean.Main.eval_bRegPreMessageExpr]
   · rw [regPreMessageExpr, ZiskFv.AirsClean.Main.eval_cRegPreMessageExpr]
 
+/-- The whole value-level register-pre message of a slot. Its value lanes are the row's **own**
+operand values, which is why matching it against `bootMessage` pins those values to `0`. -/
+def regPreMessage : RegSlot → MainRowWithRom FGL →
+    ZiskFv.Channels.MemoryBus.MemBusMessage FGL
+  | .a, row => ZiskFv.AirsClean.Main.aRegPreMessage row
+  | .b, row => ZiskFv.AirsClean.Main.bRegPreMessage row
+  | .c, row => ZiskFv.AirsClean.Main.cRegPreMessage row
+
+/-- Evaluating a slot's register-pre message gives that slot's value-level message. -/
+theorem eval_regPreMessageExpr (s : RegSlot) (env : Environment FGL)
+    (row : Var MainRowWithRom FGL) :
+    eval env (s.regPreMessageExpr row) = s.regPreMessage (eval env row) := by
+  cases s
+  · rw [regPreMessageExpr, regPreMessage, ZiskFv.AirsClean.Main.eval_aRegPreMessageExpr]
+  · rw [regPreMessageExpr, regPreMessage, ZiskFv.AirsClean.Main.eval_bRegPreMessageExpr]
+  · rw [regPreMessageExpr, regPreMessage, ZiskFv.AirsClean.Main.eval_cRegPreMessageExpr]
+
+/-- The register-pre message's own `timestamp` field is the slot's predecessor column. -/
+@[simp] theorem regPreMessage_timestamp (s : RegSlot) (row : MainRowWithRom FGL) :
+    (s.regPreMessage row).timestamp = s.prevStep row := by
+  cases s <;> rfl
+
 /-- The whole value-level read message of a slot. Keeping it as one object means a consumer that
 needs the *values* — the register telescope — does not have to re-derive them field by field. -/
 def readMessage : RegSlot → MainRowWithRom FGL →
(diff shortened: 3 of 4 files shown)


codygunton:
## Gates on `ee6f78ff`

```
lake build                            Build completed successfully (9165 jobs)
```

No `sorry`, no new trust marker. Axiom closure of the new results, all kernel-only:

```
exists_bootAnchored               [propext, Classical.choice, Quot.sound]
backward_step                     [propext, Classical.choice, Quot.sound]
backward_step_lt                  [propext, Classical.choice, Quot.sound]
bootAnchored_values_zero          [propext, Classical.choice, Quot.sound]
memBus_mem_op_three_counterpart   [propext, Classical.choice, Quot.sound]
exists_pull_of_push_memBus        [propext, Classical.choice, Quot.sound]
main_regPre_mult_zero_or_one      [propext, Classical.choice, Quot.sound]
```

**One failure worth recording rather than hiding.** The first gate run failed V1 check 18: I had not
imported `RegisterValueTelescope` from `ZiskFv.lean`, so the module was never compiled as part of the
library — every "green" module build I had been quoting was a direct `lake build <module>`, which
does not catch this. Fixed in `ee6f78ff`; the numbers above are from the re-run. That check exists
for exactly this and explicitly forbids an allowlist, which is the right call.

codygunton:
## Phase 4 stops here, and the reason is architectural

Recording this on the PR so a reviewer does not have to rediscover it.

Both halves of Phase 4 now exist. This PR is the ZisK side; the Sail side was already built and
predates it — `ZiskFv/Compliance/TraceLevelExport/RegisterWriteback.lean`,
`regs_write_of_busEffect_ok_three` and `regs_preserved_of_busEffect_ok_three`, explicitly labelled
"#330 Phase 4 groundwork".

**They cannot be composed yet.** `SailTrace n` is `Fin n → PreSail.SequentialState`
(`Compliance/SailTrace.lean:20`) — a family of states with no relation between indices. So
`RegAgree j → RegAgree (j+1)` has nothing to induct over: the writeback lemma constrains `post`, the
post-state of `bus_effect` at step `j`, while `sailTrace (j+1)` is an unrelated state.

The PC arm met the same wall and assumed the link. `SegmentPcChain` extends `SailRetireChain`, a
per-step `retire` law, and `SailRetireChain.lean:789` is blunt about the cost: *"This direction is
why Phase 7 is not, by itself, a logical-strength reduction, and the docstrings must not claim one."*

A register `SegmentRegChain.succ` would trade 116 per-row assumptions for `n` per-step assumptions of
the same logical strength. That is laundering, so I did not build it.

**Unblocker is #343** (make `SailTrace` a chained execution, so `succ` becomes definitional). After
that the composition is mechanical and the 504-occurrence strip across 35 files is bookkeeping.
Recorded in `PLAN_INPUTSAGREE_330.md` with a *do not open that premise* note.

codygunton:
## Rebased onto `main`; base retargeted from `348-boundary-anchor` to `main`

`ee6f78ff` → `0f2e48c7`, force-pushed. `348-boundary-anchor` was squash-merged as `60d08818`, so its
two commits (`f28751c5`, `e22e8f12`) were duplicated here and the PR read `CONFLICTING`. Dropped via
`git rebase --onto origin/main e22e8f12`; the eight Phase 4 commits are unchanged apart from one
import-line conflict in `ZiskFv.lean` (`main` had gained `RegisterBoundaryAnchorNonvacuity`; both
imports kept). Now **MERGEABLE**.

Gates on `0f2e48c7`: `lake build` **9166 green**, V1 **20/20**, V2 **20/20, exit 0**.

## Self-review before you merge: `exists_bootAnchored` is the WEAK shape

Flagging this myself because it is the same defect your #347 review caught, and #349 fixed there.

`backward_step` **does** carry the value link — its second disjunct ends with

```
∧ t.readMessage (…) = s.regPreMessage (…)
```

but `exists_bootAnchored_of_fuel` discards that equality when it recurses: it passes
`ih h_tbl h_comp h_r t h_tsel` and returns the endpoint's existential. So the landed statement is

```
∃ tbl ∈ …, ∃ r ∈ …, ∃ t : RegSlot, t.selector … = 1 ∧ BootAnchoredAt trace tbl r t
```

which quantifies over a **fresh** `tbl / r / t` and says nothing relating the anchor to the slot the
walk started from. Compare `exists_boundaryWalk` (#349), which records
`path.head? = some (starting row, s)` — its own docstring says to *"prefer the path version, which
says that the site you started from is the one traced back."*

So this PR proves **termination of the backward walk**, which is real content, but it does **not**
yet deliver the value telescope: `bootAnchored_values_zero` gives `value_0 = value_1 = 0` at *some*
anchor, not at the anchor of *your* slot, and no value is carried back along the chain.

I do not think that blocks merging — it mirrors #342's own history exactly, where
`exists_boundarySuppliedSite` landed first and #349 added the path form and kept the weak one as a
corollary. But **the PR description overstates it**, and I would rather correct that here than have
it read as delivered:

> "**`exists_bootAnchored`** — from any active register slot of any witness Main row, following
> register-pre pushes back reaches a slot whose predecessor access is the boot pull."

True as stated for *a* slot; what a reader will take from it — that *your* slot's value traces back —
is not proved. The path-recording version is item 0 of the Phase 4 continuation and I am building it
next, mirroring `exists_boundaryWalk_of_fuel`.

codygunton:
Closing as superseded by merged PR #356. PR #356 contains this value-telescope work plus the stronger boot-walk path, RegisterBoundary fidelity repair, and complete coverage argument.

PR #352, opened 2026-08-13
#343: root_soundness generates the Sail trace, so retire stops being a premise

### LINKED ISSUE #343: Chain SailTrace so the PC arm is genuinely discharged (protected interface — needs sign-off)

codygunton:
Closes #343.

`root_soundness` no longer takes a `SailTrace`. It takes the initial Sail state and generates the
trace itself, so the `retire` obligation stops being a premise.

**This is the first proof obligation to actually leave `root_soundness` on the #330 line.** Phases 5,
6 and 7 were restructurings — each traded one caller-supplied premise for another of the same
strength, and `SailRetireChain.lean:789` says so on the record.

## What changed

```diff
-    (sailTrace : SailTrace numInstructions)
+    (init : PreSail.SequentialState RegisterType Sail.trivialChoiceSource)
...
-    (pcChain : SegmentPcChain ziskTrace sailTrace ziskStep)
+    (pcBoot : ∀ (_ : 0 < numInstructions),
+      ((mainOfTable ziskTrace.program ziskTrace.mainTable).pc 0).val
+        = (init.regs.get? Register.PC).elim 0 BitVec.toNat)
```

`chainedSailTrace` runs `execute_instruction` at each step and then performs Sail's retire — copy
`nextPC` into `PC`. `SegmentPcChain`'s `retire` field is supplied inside the theorem by
`chainedSailTrace_retireChain`, which takes **no hypothesis**.

**Be precise about what left.** The binder *count* is unchanged at 16: `pcChain` was **replaced** by
`pcBoot`, not removed. What left is `SegmentPcChain`'s `retire` field. And the edit does two things —
it drops that obligation *and* specializes the conclusion, since the new statement is an **instance**
of the old one at `sailTrace := chainedSailTrace ziskStep init`. A caller who wants `StepSound` at a
trace of their own choosing can no longer get it here. That is the intended direction; it is a
specialization, not a free strengthening.

One detail that made `retire` unconditional: `retirePC`'s `nextPC`-absent branch **erases** `PC`, so
both sides of the law are `none` there. Writing it the obvious way instead leaves a side condition
that would need a fact about all 63 `execute` arms. That is a coverage cliff, not a soundness hole —
`h_input_pc` demands `.some`, so an erased `PC` makes `inputsAgree` unsatisfiable rather than
silently matching `.elim 0`'s zero.

## Why this is a reduction and not another restructure

`sailRetireChain_of_inputsAgree` derives `retire` from the old per-row bundle. That converse is
exactly what made Phase 7 inter-derivable with what it replaced, and it needs `retire` to be
something a caller supplies. At the root there is nothing to supply, so there is nothing to derive.

## No witness exercises a chain step — stated, not buried

Both instantiations run at `numInstructions ≤ 1`, and `chainedSailStates … 0` is `init` by `rfl`, so
index `0` is the caller's own state. `addFaithfulInputsAgree` is textually unchanged by this PR —
only `addFaithfulInputsAgreeCore` moved, and only by gaining `noncomputable`. So **`retirePC` and
`sailStepPost` are never applied anywhere in the tree, and `chainedSailTrace_retireChain` is proved
but never used by a witness.**

The premise reduction holds regardless — it is a fact about the statement. What is *not* yet shown is
an instantiation that had to run Sail to satisfy `inputsAgree`, i.e. #343's "harder to inhabit"
guardrail demonstrated rather than argued. This is now said in both `ChainedSailTrace.lean` and
`trust/trusted-base.md`, next to the reduction claim rather than in a separate paragraph.

## The scope claim in #343 was wrong

#343 says `SailTrace` is referenced by all 63 `StepSound` arms, every `Inputs_<op>`, and all 7
witnesses, and therefore needs sign-off before starting. Against the tree:

- `InputsAgreeCore` and `StepSoundWithoutDecode` take `sailTrace` as a **parameter** and only ever
  use `sailTrace i`. The 63 arms need no edit. There are 11 `sailTrace` binder sites in total, 8 of
  them in `Dispatcher.lean`.
- Only **2** of the 7 witnesses instantiate `root_soundness`, at `n = 1` and `n = 0`. The other 5
  target `stepSound_of_programDecodes`, which deliberately keeps its signature and is untouched.

I asserted the larger figure from the issue body rather than the tree, and it is the reason this work
looked infeasible for longer than it was.

## What this does **not** do

`stepSound_of_programDecodes` still takes a `SailTrace` and a full `SegmentPcChain`. The five
multi-step witnesses (`addSpin`, `addAddiSpin`, `divSpin`, `jalrSpin`, `sdLdSpin`) keep their
hand-written state families. Migrating them means deriving each state by running Sail — 13 chain
steps — and that is what would exercise the chain. Not claimed here. The route, if picked up, is not
brute-force reduction of Sail's `execute`: the `Dispatch/` equivalences already give
`execute_instruction … = state_effect_via_channels …`, so the post-state is computable from each
witness's own bus rows via `nextPC_of_busEffect_ok` and `RegisterWriteback`'s projections.

## V2 check 6 was an implementation bug in the gate, not a policy question

The first push had check 6 failing on the two binders that mention `chainedSailTrace`. I proposed a
carve-out. **That was the wrong fix, and the review was right to reject it.**

`bin/TrustGate/TypeWalk.lean`'s `shouldUnfold` tested `ConstantInfo.hints`, treating
`ReducibilityHints.regular n` with `n > 0` as reducible. `regular n` carries a definition's *height*,
which is positive for ordinary `def`s too — so the walk unfolded essentially **every** non-recursive
definition, far wider than the module's documented contract (*"`abbrev` / `@[reducible] def`
chains"*). It now asks `getReducibilityStatusCore`. Both binders clear with
`trust/forbidden-types.txt` **untouched**, and check 2 still passes on all 63 canonical `equiv_<OP>`
theorems.

This is a coverage reduction, and it is recorded as one in `trust/trusted-base.md` ("TypeWalk
coverage"). The boundary is measured, not asserted:

| declaration | status | walk |
|---|---|---|
| `abbrev SneakySpec := LeanRV64D.Functions.execute` | reducible | unfolds — **dodge still closed** |
| `def SneakyOrdinary := LeanRV64D.Functions.execute` | semireducible | stops — **newly missed** |
| `ZiskFv.Compliance.SailTrace` (`abbrev`) | reducible | unfolds |
| `chainedSailTrace` / `chainedSailStates` | semireducible | stops |

So the aliasing dodge this layer exists to close stays closed, and direct syntactic occurrences never
needed unfolding at all. What is newly missed is an *ordinary-`def`* alias of a Sail entry point in a
binder type. The old behaviour over-flagged and never under-flagged, so no past result is
invalidated. Both docstrings were corrected — the gap between stated and actual behaviour is what
turned a bug into an apparent policy question.

`/bin/TrustGate/` and `/trust/trusted-base.md` are both CODEOWNER-protected, so this needs your ack
even though the gate is now green.

## Trust boundary

No new `axiom`, `opaque`, `sorry`, `native_decide`, or any other trust marker. No allowlist edit, and
`trust/forbidden-types.txt` is unchanged.

The new declarations close over `[propext, Classical.choice, Quot.sound]`; those mentioning
`execute_instruction` close over the pre-existing Sail base (`cancel_reservation`, `riscv_f*`, …)
that the endpoint already carried. Both witness axiom closures are unchanged.

Frozen artefacts changed **together with** the declaration, as the trust workflow requires:

- `ZiskFv/Audit.lean` — the `#guard_msgs` statement snapshot;
- `trust/generated/baseline-strong-export-binders.txt` — regenerated (16 entries, before and after);
- `trust/trusted-base.md` — the PC-arm section, whose paragraph headed *"What would be a real
  strength reduction, and is not done"* described precisely this change, plus the new "TypeWalk
  coverage" note.

## Verification

`lake build` **9166 green**, V1 **20/20**, V2 **20/20**. Logs in the comment below.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -198,6 +198,7 @@ import ZiskFv.Compliance.RegisterWalk
 import ZiskFv.Compliance.MemBusSlotSeparation
 import ZiskFv.Compliance.RegisterBoundaryAnchor
 import ZiskFv.Compliance.RegisterBoundaryAnchorNonvacuity
+import ZiskFv.Compliance.TraceLevelExport.ChainedSailTrace
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Compliance.AddSpinRootSoundness
--- a/ZiskFv/Audit.lean
+++ b/ZiskFv/Audit.lean
@@ -65,21 +65,23 @@ deliberately changes what an endpoint states. -/
 open ZiskFv.Compliance in
 /--
 info: root_soundness : ∀ (numInstructions rawLength : ℕ) (ziskTrace : AcceptedZiskTrace numInstructions)
-  (sailTrace : SailTrace numInstructions) (ziskStep : (i : Fin numInstructions) → ZiskStep ziskTrace i)
-  (start : Fin rawLength → Fin ziskTrace.programLength) (addr : Fin rawLength → Fin 18446744069414584321)
-  (rawProgram : Fin rawLength → BitVec 32)
+  (init : PreSail.SequentialState RegisterType Sail.trivialChoiceSource)
+  (ziskStep : (i : Fin numInstructions) → ZiskStep ziskTrace i) (start : Fin rawLength → Fin ziskTrace.programLength)
+  (addr : Fin rawLength → Fin 18446744069414584321) (rawProgram : Fin rawLength → BitVec 32)
   (programBinding : RawProgramBinding.ProgramRowsBinding ziskTrace start addr rawProgram)
   (rawProgramDecodes : (i : Fin numInstructions) → RawProgramDecode ziskTrace i (ziskStep i) start addr rawProgram)
-  (inputsAgree : (i : Fin numInstructions) → InputsAgreeCore ziskTrace sailTrace i (ziskStep i)),
-  SegmentPcChain ziskTrace sailTrace ziskStep →
+  (inputsAgree : (i : Fin numInstructions) → InputsAgreeCore ziskTrace (chainedSailTrace ziskStep init) i (ziskStep i)),
+  (0 < numInstructions →
+      ↑((ZiskFv.AirsClean.FullEnsemble.mainOfTable ziskTrace.program ziskTrace.mainTable).pc 0) =
+        (init.regs.get? Register.PC).elim 0 BitVec.toNat) →
     (StepRowsAligned ziskTrace ziskStep fun i ↦
         rowDecode_of_programDecode ziskTrace i
           (programDecode_of_rawProgramDecode ziskTrace i (ziskStep i) start addr rawProgram programBinding
             (rawProgramDecodes i))) →
-      ∀ (bootSeed : BootSegmentMemorySeed ziskTrace sailTrace ziskStep),
+      ∀ (bootSeed : BootSegmentMemorySeed ziskTrace (chainedSailTrace ziskStep init) ziskStep),
         (∀ (i : Fin numInstructions), RowOutsideDefectRegion ziskTrace i (ziskStep i)) →
           ∀ (i : Fin numInstructions),
-            StepSound ziskTrace sailTrace i (ziskStep i)
+            StepSound ziskTrace (chainedSailTrace ziskStep init) i (ziskStep i)
               (rowDecode_of_programDecode ziskTrace i
                 (programDecode_of_rawProgramDecode ziskTrace i (ziskStep i) start addr rawProgram programBinding
                   (rawProgramDecodes i)))
--- a/ZiskFv/Compliance/AddFaithfulPaddedRootSoundness.lean
+++ b/ZiskFv/Compliance/AddFaithfulPaddedRootSoundness.lean
@@ -54,8 +54,12 @@ def addFaithfulState (pc : BitVec 64) :
     regs := addFaithfulRegs pc
     mem := {} }
 
-def addFaithfulSailTrace : SailTrace 1
-  | ⟨0, _⟩ => addFaithfulState (0#64)
+/-- The Sail side of this witness (#343). Not a hand-written family of states: the execution Sail's
+    own semantics generates from `addFaithfulState (0#64)`. A one-instruction execution has no chain
+    step, so it reduces to the initial state at its single index — but it is now the *generated*
+    trace, which is what `root_soundness` consumes. -/
+noncomputable def addFaithfulSailTrace : SailTrace 1 :=
+  chainedSailTrace addFaithfulZiskStep (addFaithfulState (0#64))
 
 theorem addFaithfulRowsOf_empty_readSound :
     MemoryBusRowsPrefixReadSound
@@ -205,7 +209,7 @@ def addFaithfulInputsAgree :
 def addFaithfulPcSeed : SegmentPcSeed addFaithfulAcceptedTrace addFaithfulSailTrace :=
   pcSeed_of_inputsAgree addFaithfulInputsAgree
 
-def addFaithfulInputsAgreeCore :
+noncomputable def addFaithfulInputsAgreeCore :
     ∀ i : Fin 1,
       InputsAgreeCore addFaithfulAcceptedTrace addFaithfulSailTrace i (addFaithfulZiskStep i) :=
   fun i => inputsAgreeCore_of_inputsAgree i (addFaithfulZiskStep i) (addFaithfulInputsAgree i)
@@ -267,15 +271,14 @@ def addFaithfulRowsAligned :
   have h' : j + 1 < 1 := h
   omega
 
-/-- The two root PC premises: boot agreement, and the Sail-internal retire law. `retire` is vacuous
-    on a one-instruction execution; `boot` is the witness's own row-0 PC agreement. -/
-def addFaithfulPcChain :
-    SegmentPcChain addFaithfulAcceptedTrace addFaithfulSailTrace addFaithfulZiskStep where
-  retire := by
-    intro j h
-    have h' : j + 1 < 1 := h
-    omega
-  boot := (pcSeed_of_inputsAgree addFaithfulInputsAgree).boot
+/-- The single root PC premise (#343): boot agreement at row `0`. The retire law that used to sit
+    beside it is no longer a premise — `root_soundness` gets it from
+    `chainedSailTrace_retireChain`. -/
+def addFaithfulPcBoot : ∀ (_ : 0 < 1),
+    ((ZiskFv.AirsClean.FullEnsemble.mainOfTable addFaithfulAcceptedTrace.program
+        addFaithfulAcceptedTrace.mainTable).pc 0).val
+      = ((addFaithfulState (0#64)).regs.get? Register.PC).elim 0 BitVec.toNat :=
+  (pcSeed_of_inputsAgree addFaithfulInputsAgree).boot
 
 open ZiskFv.Compliance.RawProgramBinding in
 theorem addFaithfulPaddedRawRootSoundness :
@@ -285,9 +288,9 @@ theorem addFaithfulPaddedRawRootSoundness :
           (programDecode_of_rawProgramDecode addFaithfulAcceptedTrace i (addFaithfulZiskStep i)
             addFaithfulStart addFaithfulAddr addFaithfulRawProgram
             addFaithfulProgramRowsBinding (addFaithfulRawProgramDecodes i))) :=
-  root_soundness 1 2 addFaithfulAcceptedTrace addFaithfulSailTrace addFaithfulZiskStep
+  root_soundness 1 2 addFaithfulAcceptedTrace (addFaithfulState (0#64)) addFaithfulZiskStep
     addFaithfulStart addFaithfulAddr addFaithfulRawProgram addFaithfulProgramRowsBinding
-    addFaithfulRawProgramDecodes addFaithfulInputsAgreeCore addFaithfulPcChain
+    addFaithfulRawProgramDecodes addFaithfulInputsAgreeCore addFaithfulPcBoot
     addFaithfulRowsAligned addFaithfulBootSeed addFaithfulOutsideDefectRegion
 
 #print axioms addFaithfulPaddedRawRootSoundness
--- a/ZiskFv/Compliance/MemoryRawRootSoundness.lean
+++ b/ZiskFv/Compliance/MemoryRawRootSoundness.lean
@@ -77,11 +77,17 @@ private theorem memoryAcceptedTrace_not_mutableMemPresent :
   have htable := memoryAcceptedTrace_mutable_tables_empty table hmem hcomp
   exact absurd hlen (by simp [htable])
 
-/-- The empty Sail trace over the empty execution. -/
-def memorySailTrace : SailTrace 0 := nofun
-
 /-- The empty per-step ZisK decode family over the empty execution. -/
 def memoryZiskStep : ∀ i : Fin 0, ZiskStep memoryAcceptedTrace i := nofun
+
+/-- The initial Sail state this witness hands to `root_soundness` (#343). An empty execution never
+    reads it, but the theorem now takes a state rather than a trace, so it must be supplied. -/
+def memoryInit : PreSail.SequentialState RegisterType Sail.trivialChoiceSource := default
+
+/-- The Sail trace over the empty execution — generated from `memoryInit`, not hand-written. Over
+    `Fin 0` it has no inhabited index, as before. -/
+noncomputable def memorySailTrace : SailTrace 0 :=
+  chainedSailTrace memoryZiskStep memoryInit
 
 /-- The boot / cross-segment memory seed for the memory witness: empty boot memory, no
     execution-order rows, and every per-index obligation either vacuous over `Fin 0`
@@ -100,12 +106,14 @@ def memoryBootSeed :
     exact absurd (by simp [AcceptedZiskTrace.numInstructions]) h_ne
   placement := fun i => i.elim0
 
-/-- The PC premises for the empty execution. All of them are vacuous here for the same
-    reason `memoryBootSeed`'s are: there is no step `0`, and no step `j + 1`. -/
-def memoryPcChain :
-    SegmentPcChain memoryAcceptedTrace memorySailTrace memoryZiskStep where
-  retire := fun _ h => absurd h (Nat.not_lt_zero _)
-  boot := fun h => absurd h (Nat.not_lt_zero _)
+/-- The one PC premise left after #343, vacuous here for the same reason `memoryBootSeed`'s fields
+    are: the empty execution has no step `0`. The retire law that used to sit beside it is no longer
+    a premise at all — `root_soundness` derives it from `chainedSailTrace_retireChain`. -/
+def memoryPcBoot : ∀ (_ : 0 < 0),
+    ((ZiskFv.AirsClean.FullEnsemble.mainOfTable memoryAcceptedTrace.program
+        memoryAcceptedTrace.mainTable).pc 0).val
+      = (memoryInit.regs.get? Register.PC).elim 0 BitVec.toNat :=
+  fun h => absurd h (Nat.not_lt_zero _)
 
 def memoryRowsAligned :
     StepRowsAligned memoryAcceptedTrace memoryZiskStep
@@ -125,9 +133,9 @@ theorem memoryRawRootSoundness :
         (rowDecode_of_programDecode memoryAcceptedTrace i
           (programDecode_of_rawProgramDecode memoryAcceptedTrace i (memoryZiskStep i)
             memoryStart memoryAddr memoryRawProgram memoryProgramRowsBinding i.elim0)) :=
-  root_soundness 0 3 memoryAcceptedTrace memorySailTrace memoryZiskStep
+  root_soundness 0 3 memoryAcceptedTrace memoryInit memoryZiskStep
     memoryStart memoryAddr memoryRawProgram memoryProgramRowsBinding
-    (fun i => i.elim0) (fun i => i.elim0) memoryPcChain memoryRowsAligned memoryBootSeed
+    (fun i => i.elim0) (fun i => i.elim0) memoryPcBoot memoryRowsAligned memoryBootSeed
     (fun i => i.elim0)
 
 #print axioms memoryRawRootSoundness
--- a/ZiskFv/Compliance/TraceLevelExport/ChainedSailTrace.lean
+++ b/ZiskFv/Compliance/TraceLevelExport/ChainedSailTrace.lean
@@ -17,12 +17,19 @@ This module builds the trace instead of assuming it. `chainedSailStates` runs
 
 It settles the shape: a trace generated this way satisfies `retire` definitionally, so a caller that
 supplies one is supplying an actual Sail execution rather than an arbitrary family of states with a
-promise attached.
+promise attached. `root_soundness` consumes it — that theorem takes an initial state, builds the
+trace here, and no longer carries a `SegmentPcChain` binder.
 
-It does not by itself migrate `root_soundness`. `SailTrace` is referenced by all 63 `StepSound`
-arms, every `Inputs_<op>`, and all seven accepted-trace witnesses; repointing them is the rest of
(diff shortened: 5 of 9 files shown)


codygunton:
## Gates on `73130cb1` (branch cut from `origin/main` = `60d08818`)

| gate | result |
|---|---|
| `lake build` | **green, 9166 jobs** |
| `trust/scripts/check-all.sh` (V1) | **20/20, exit 0** |
| `trust/scripts/check-all-semantic.sh` (V2) | **19/20, exit 1** — check 6 only, see the PR body |
| `sorry` / `admit` under `ZiskFv/` | none added |
| new `axiom` / `opaque` / `native_decide` | none |

V2 detail — every other check passes:

```
1/20  per-theorem axiom-dep baseline matches.
2/20  no forbidden types in 63 canonical equiv_<OP> theorem(s).      <- the catalog's home scope
3/20  uber-theorem axiom closure matches baseline-axioms.txt.
4/20  global theorem binder baseline matches.
5/20  strong-export axiom closure matches baseline (0 ZiskFv.* axiom(s)).
6/20  strong-export binders: BASELINE MATCHES, forbidden-Name walk FAILS on #10 and #13.
7..20 pass (incl. 16/20 root_soundness instantiation witnesses,
      19/20 register walk, 20/20 3017 extraction/decode raw closures).
```

Check 5 is the one worth reading next to check 6: the **axiom** closure is unchanged at 0 `ZiskFv.*`
axioms. Nothing entered the trust boundary; what changed is that a premise's *type* now names Sail's
step function, because the state it talks about is defined by running Sail.

## Axiom closures of the two migrated witnesses

Both unchanged from before the migration — the pre-existing Sail base reached through
`execute_instruction`, which `root_soundness` already carried:

```
'ZiskFv.Compliance.AddFaithfulPaddedRootSoundness.addFaithfulPaddedRawRootSoundness'
  depends on axioms: [cancel_reservation, get_16_random_bits, load_reservation, match_reservation,
                      plat_term_write, propext, riscv_f16Add, ... , Classical.choice, Quot.sound]
'ZiskFv.Compliance.MemoryRawRootSoundness.memoryRawRootSoundness'
  depends on axioms: [cancel_reservation, ...]
```

The new declarations in `ChainedSailTrace.lean` that do not mention `execute_instruction`
(`retirePC_regs_of_ne`, `chainedSailStates_zero`, `chainedSailTrace_apply`) close over
`[propext, Classical.choice, Quot.sound]`.

## Reviewer's shortcut

The whole change is legible from one diff hunk, `ZiskFv/Soundness.lean`:

- `sailTrace : SailTrace numInstructions` → `init : PreSail.SequentialState …`
- `pcChain : SegmentPcChain ziskTrace sailTrace ziskStep` → `pcBoot`, the boot equation alone
- the proof term gains
  `{ toSailRetireChain := chainedSailTrace_retireChain ziskStep init, boot := pcBoot }`

`stepSound_of_programDecodes` directly below it is unchanged, which is what keeps the five
multi-step witnesses out of this PR.

codygunton:
## Review of `73130cb1`

Reproduced independently in a worktree detached at the PR head:

```
lake build                            Build completed successfully (9166 jobs)
trust/scripts/check-all-semantic.sh   19/20, check 6 the only failure
```

Your numbers reproduce exactly. Within check 6, `strong-export binder baseline matches` — the
regenerated baseline is fine; only the forbidden-Name walk fails.

**Verdict: the reduction is real and the Lean is sound. But the check-6 failure is a bug in the
gate, not a policy question — so the carve-out should not be granted. And the one property that
makes this a reduction rather than a restructure is not exercised by any witness.**

---

# 1. Check 6: the gate is misbehaving, and the carve-out would paper over it

**Do not grant the carve-out.** The walk is not detecting an aliasing dodge. It is unfolding
definitions it was never meant to unfold.

`bin/TrustGate/TypeWalk.lean::shouldUnfold` decides what to unfold like this:

```lean
| .regular n => n > 0  -- @[reducible] sets a positive depth hint
```

That comment is false. `@[reducible]` sets an *attribute*; `ReducibilityHints.regular n` is an
unfolding **height**, and ordinary definitions get positive heights too. Measured in this tree:

```
MyAbbrev                  :: hints=abbrev     :: attr=REDUCIBLE
MyReducible (@[reducible] def) :: hints=regular 2  :: attr=REDUCIBLE
MyPlain     (plain def)        :: hints=regular 2  :: attr=semireducible
chainedSailTrace               :: hints=regular 49 :: attr=semireducible
sailStepPost                   :: hints=regular 47 :: attr=semireducible
```

A `@[reducible] def` and a plain `def` are **indistinguishable** by the hint. So the predicate
returns `true` for essentially every ordinary definition, and the walk is a near-total transitive
constant closure that stops only at structures, inductives and `opaque`. That is far broader than
what both docstrings promise — the file header says "closure under reducible definitions", the
catalog header says "after reducible-transparency unfolding".

That is the whole mechanism behind the failure: `chainedSailTrace` (semireducible) is unfolded, its
body reaches `sailStepPost` (semireducible), which syntactically contains `execute_instruction`.

I re-ran both closures over the two flagged binder types with the predicate corrected to test the
actual reducibility attribute:

```
binder #10:
  current predicate    -> hits: [execute_instruction, LeanRV64D.Functions.execute]
  reducible-attr only  -> hits: []
binder #13:
  current predicate    -> hits: [execute_instruction, LeanRV64D.Functions.execute]
  reducible-attr only  -> hits: []
```

**So the fix is a one-line correction in `shouldUnfold`, not a carve-out**: test
`getReducibilityStatusCore env n` instead of the hint. Check 6 then passes with
`trust/forbidden-types.txt` untouched, no policy exception, and nothing marked irreducible.

Three things to note about that fix:

- **It does not reopen the dodge the layer exists to close.** `abbrev S := execute` and
  `@[reducible] def S := execute` both report `REDUCIBLE` above, so both stay unfolded.
- **It narrows the walk to its documented contract**, so it is a deliberate reduction in coverage
  and belongs in the ledger as one. No past result is invalidated: the walk has been
  *over*-flagging, never under-flagging, so it never produced a false pass.
- **Both docstrings should be corrected too** — they describe the intent, not the behaviour, and
  that gap is what turned an implementation bug into a policy question.

Also, on the carve-out as drafted: "admit a Sail reference reaching a binder **only** through
`chainedSailTrace`" is open-ended. It would admit *any* future binder mentioning the generated
trace, including one of genuine OUTPUT-EQ shape. If you do decide to keep the broad walk and carve
out instead, pin the exception to the two binder entries in
`trust/generated/baseline-strong-export-binders.txt` — that file is CODEOWNER-protected under
`/trust/`, so an exact-binder pin is drift-safe, whereas a rule keyed on `chainedSailTrace` is not.

Your OUTPUT-EQ argument is separately correct, for the record: a premise that is strictly *harder*
to satisfy cannot hand the theorem its conclusion. The catalog's inference — "Sail entry point in a
parameter ⇒ OUTPUT-EQ" — genuinely stops being valid once the trace is spec-generated. I just don't
think you need it, because the gate should not have fired at all.

---

# 2. No witness exercises a chain step

This is the finding I would not merge without addressing, at least in the ledger.

`root_soundness` has two instantiations, at `numInstructions = 1` and `0`. `chainedSailStates … 0 =
init` is `rfl`, so at `n = 1` the only index reduces to the caller's own initial state:

- `addFaithfulSailTrace` is now `chainedSailTrace addFaithfulZiskStep (addFaithfulState (0#64))`,
  but `addFaithfulInputsAgree` is **unchanged** — it is not even in the diff. Only
  `addFaithfulInputsAgreeCore` changed, and only by gaining `noncomputable`.
- `memoryRawRootSoundness` is `Fin 0`; every premise is vacuous.

So `retirePC` and `sailStepPost` are never applied at any index of any checked-in instantiation, and
`chainedSailTrace_retireChain` is proved but never exercised.

That matters because the guardrail is the entire argument for calling this a reduction:

> **The guardrail #343 asks for is met.** A caller can no longer choose the Sail states. […] The
> trace became harder to inhabit rather than merely repackaged — which is the test that separates
> this from the Phase 5/6/7 restructurings above.

For both witnesses the caller still chooses the Sail state completely, because index `0` *is* the
caller's `init`. "Harder to inhabit" is true of the statement and untested in the tree. The ledger
states the guardrail as met and separately reports the `n = 1` / `n = 0` non-vacuity, but never
connects the two.

I am not asking for the five-witness migration — you scoped that out and I agree with the scoping.
I am asking that the ledger say plainly that no witness reaches index `≥ 1`, so the chain link is
unexercised. A witness at `n = 2` would be the real answer; short of that, the honest sentence is
cheap and keeps the claim calibrated.

---

# 3. `sailStepPost`'s docstring contradicts its code

```lean
/-- The post-state of one Sail step, error branches kept total by staying put. -/
  | .ok _ post => post
  | .error _ post => post
```

Both branches return `post`. The error branch does not stay put — it takes the **trap's**
post-state and the chain advances from there. So the generated trace continues through a Sail trap
as though it were a normal step, and `chainedSailTrace_retireChain` says nothing about that index
(its law is conditioned on `.ok`). Either correct the docstring and say what a trapping step means
for the generated trace, or make the branch return `s`.

# 4. `retirePC`'s erase branch is safe, but it is a coverage cliff

Worth recording because the reasoning is not obvious. `InputsCore_*.h_input_pc` demands
`(binding i).regs.get? Register.PC = .some …`, so a PC-erased state makes `inputsAgree`
**unsatisfiable** at that index rather than silently matching `boot`'s `.elim 0` zero. No soundness
hole — good.

The cost is instantiability: any step whose `execute` leaves `nextPC` unset makes the theorem
uninstantiable from that index onward. That is exactly what the "fact about all 63 `execute` arms"
would have ruled out. The ledger records the mechanism ("erases `PC`, so both sides are `none`") but
not this consequence. One line.

# 5. The binder count did not change — 13 before, 13 after

`baseline-strong-export-binders.txt` shows `11 :: pcChain :: SegmentPcChain …` replaced by
`11 :: pcBoot :: …`, not removed. What left the endpoint is the `retire` **field** of that
structure, which is a genuine weakening of the binder and a real reduction. But "the `pcChain`
binder is gone" and "the first premise to actually leave the endpoint" both read as a shorter
premise list, and the audit surface says otherwise. Say "the binder weakened from
`SegmentPcChain` to its `boot` half".

Related framing point: the new `root_soundness` is an **instance** of the old one — take
`sailTrace := chainedSailTrace ziskStep init` and discharge `retire`. So the edit both removes an
obligation and specializes the statement to chained traces. The specialization is the right
direction for a soundness endpoint and I would not want it otherwise, but the ledger currently
states only the first half.

---

## What I checked and agree with

- `pcBoot` is exactly the old `boot` field with `binding ⟨0,h⟩` specialized to `init`, including the
  `.elim 0 BitVec.toNat` convention. No weakening smuggled in there.
- `chainedSailTrace_retireChain` really is hypothesis-free, and the `none`/`some` case split is what
  makes it so.
- `Audit.lean`, the binder baseline and `trust/trusted-base.md` all moved with the declaration, as
  the trust workflow requires.
- The corrected scope claim in the PR body matches the tree: `InputsAgreeCore` takes `sailTrace` as a
  parameter and the 63 arms need no edit.

codygunton:
## Review round 1 applied — all five findings upheld, gates now fully green

`466e7663`. Thank you for the check-6 diagnosis; correcting `shouldUnfold` was clearly the right call
and my carve-out proposal was not.

| gate | before (`73130cb1`) | after (`466e7663`) |
|---|---|---|
| `lake build` | green, 9166 | **green, 9166** |
| V1 `check-all.sh` | 20/20, exit 0 | **20/20, exit 0** |
| V2 `check-all-semantic.sh` | 19/20, exit 1 | **20/20, exit 0** |

```
::: 6/20 strong-export binders + forbidden types (V2) :::
trust-gate (V2): strong-export binder baseline matches.
trust-gate (V2): no forbidden types in ZiskFv.Compliance.root_soundness binders.
...
trust-gate (V2 semantic): ALL CHECKS PASSED.
```

Zero forbidden-Name hits anywhere in the log, with `trust/forbidden-types.txt` untouched.

### 1 — gate fixed, not carved out

`shouldUnfold` now asks `getReducibilityStatusCore`. I took your primary option rather than the
binder-baseline pin; your reasoning that the pin would be drift-safe is right, but the pin preserves a
walk whose behaviour does not match its own docstring, and that mismatch is what cost the round.

Both riders done. The coverage reduction is recorded in `trust/trusted-base.md` under **"TypeWalk
coverage — deliberate narrowing, #343"**, and I measured the boundary rather than asserting it —
`run_cmd` over `getReducibilityStatusCore` on planted aliases:

| declaration | status | walk |
|---|---|---|
| `abbrev SneakySpec := LeanRV64D.Functions.execute` | reducible | unfolds — dodge still closed |
| `def SneakyOrdinary := LeanRV64D.Functions.execute` | semireducible | stops — **newly missed** |
| `ZiskFv.Compliance.SailTrace` (`abbrev`) | reducible | unfolds |
| `chainedSailTrace` / `chainedSailStates` | semireducible | stops |

So the newly-missed case is precisely an ordinary-`def` alias of a Sail entry point. Over-flagging
only, never under-flagging, so no past result is invalidated. Both docstrings corrected, including the
module header, which now states the scope limit explicitly instead of describing a narrower walk than
the code performed.

### 2 — no witness reaches index ≥ 1

Stated in both places, and next to the reduction claim rather than in a separate paragraph:
`ChainedSailTrace.lean`'s header and `trusted-base.md`'s bullet on the guardrail. The wording is that
`retirePC` and `sailStepPost` are never applied in the checked-in tree and
`chainedSailTrace_retireChain` is proved but never used by a witness; the premise reduction holds as a
statement-level fact, and "harder to inhabit" is so far argued rather than demonstrated.

You were right that `addFaithfulInputsAgree` is not in the diff. Only `addFaithfulInputsAgreeCore`
moved, and only by gaining `noncomputable`.

### 3 — `sailStepPost` docstring

Corrected. Both branches return `post`, so a trapping step advances the chain from the trap's
post-state; the retire law is silent there because `chainedSailTrace_retireChain`'s hypothesis is an
`.ok`. I left the behaviour alone and fixed the description.

### 4 — `retirePC` erase branch

Documented on the definition and in the ledger as a coverage cliff with your reasoning: `h_input_pc`
demands `.some`, so an erased `PC` makes `inputsAgree` unsatisfiable rather than silently matching
`.elim 0`'s zero — no soundness hole, but such a trace is lost, which is what the 63-arm fact would
have ruled out.

### 5 — framing

Corrected in the theorem docstring, the ledger, and the PR body. The binder count is unchanged at 16;
`pcChain` was replaced by `pcBoot` and what left is `SegmentPcChain`'s `retire` field. Also recorded
that the new statement is an **instance** of the old one, so the edit both drops an obligation and
specializes to chained traces — right direction, but a specialization, not a free strengthening.

### Still needs your ack

`/bin/TrustGate/TypeWalk.lean` and `/trust/trusted-base.md` are CODEOWNER-protected. The gate is green
now, but the enforcement logic changed, so this is not a self-approving green.

codygunton:
## Re-review of `466e7663`

Reproduced independently, worktree detached at the updated head:

```
lake build                            Build completed successfully (9166 jobs)

  ::: 2/20 forbidden types (V2) :::
  no forbidden types in 63 canonical equiv_<OP> theorem(s).
  ::: 6/20 strong-export binders + forbidden types (V2) :::
  strong-export binder baseline matches.
  no forbidden types in ZiskFv.Compliance.root_soundness binders.
```

**All five findings are addressed. The gate is green without a carve-out and without touching
`trust/forbidden-types.txt`. Check 2 still covers all 63 arms after the narrowing — that was the
risk of the fix, and it is measured, not assumed.**

### The coverage table is exact

I re-ran the closure with the landed predicate rather than taking the table on trust:

```
SneakySpec      (abbrev, reducible)    :: forbidden reachable = true    ← dodge still closed
SneakyOrdinary  (plain def, semired.)  :: forbidden reachable = false   ← newly missed
SailTrace       (abbrev, reducible)    :: forbidden reachable = false
chainedSailTrace / chainedSailStates   :: forbidden reachable = false   ← stops, as intended
```

Every row matches what the ledger claims.

### One residual, offered as follow-up rather than a blocker

`def SneakyOrdinary := LeanRV64D.Functions.execute` in a binder type now passes both check 2 and
check 6. That is the same aliasing class the V2 layer exists to close, reached through `def` instead
of `abbrev`. The ledger names it plainly, which is the right call, and it is strictly narrower than
the pre-fix behaviour was wrong — but it is a hole, and it is closeable without giving back the
over-flagging.

The distinction you want is *alias* versus *construction*. `def S := execute` and
`def S x y := execute x y` are aliases; `chainedSailStates` is a construction that calls `execute`.
A predicate that separates them: unfold if the declaration is reducible, **or** if its value, after
stripping lambdas and `mdata`, is a constant or an application whose head is a constant and whose
arguments are all bound variables. `chainedSailStates` does not match — its body is a
`WellFounded.fix` / `brecOn` application carrying non-`bvar` arguments.

If you would rather keep `TypeWalk` semantics untouched, the decoupled alternative is a separate
whole-environment check: assert that no declaration under `ZiskFv/` is a delta-alias of a Name in
`trust/forbidden-types.txt`, independent of any binder. Either closes the dodge; neither reopens the
over-flagging.

### What landed, against the five findings

1. **Check 6.** `shouldUnfold` now takes `env` and tests `getReducibilityStatusCore`. Both docstrings
   corrected, and the module header states the scope rather than the intent — including the sentence
   that the previous wording "did not match the code and that mismatch cost a review round."
2. **Chain step unexercised.** Recorded in three places: the module header, the `root_soundness`
   docstring, and the ledger, which now separates "met at the level of the statement" from "NOT yet
   demonstrated by a witness" and names the evidence — `addFaithfulInputsAgree` is textually
   unchanged by the migration.
3. **`sailStepPost`.** Docstring corrected, and it says outright that the earlier one was wrong.
4. **`retirePC` erase branch.** Now carries the coverage-cliff reasoning, including that
   `h_input_pc` demands `.some`, so the failure mode is uninstantiability rather than a silent
   `.elim 0` match.
5. **Binder count.** Corrected with the file citation. My "13" was wrong — the baseline is 0-indexed
   and I counted only the range visible in the diff. It is 16, unchanged before and after, and what
   left is `SegmentPcChain`'s `retire` field.

The specialization point is also now on the record in both the docstring and the ledger: the new
statement is an instance of the old one, so the edit drops an obligation *and* narrows the
conclusion to chained traces.

### Verdict

Mergeable. The remaining gap — no instantiation that has to run Sail — is documented in the three
places a reader would look, and closing it is the five-witness migration you scoped out, not
something this PR should carry.

Issue #355, opened 2026-08-17
RegisterBoundary modelled the compile-time boot loop with a free reg column (model-fidelity defect)


codygunton:
## The defect

`ZiskFv/AirsClean/RegisterBoundary.lean` modelled `RegisterBoundaryRow.reg` as a **free witness
column**, and `RegisterBoundary.circuit` had `Spec := fun _ _ _ => True`. The component asserted no
algebraic relation at all: its `main` was two `MemBusChannel.emit` calls and nothing else.

Therefore the model permitted a witness with **two boundary rows that carry the same `reg`**. Two
boundary rows for one register give two boot pulls at `(3, r, 0, 0)`. Two Main rows can then anchor
at boot for `r`. That register's accesses can split into two disjoint chains.

## Why the model was wrong

ZisK does not permit that. `main.pil:536` emits the boot pulls from a compile-time loop:

```
for (int ireg = 0; ireg < REGS_IN_MAIN; ++ireg) {
    global_init_mem(sel: 1, addr: ireg + REGS_IN_MAIN_FROM, value: zeros);
}
```

`REGS_IN_MAIN_FROM = 1` and `REGS_IN_MAIN_TO = 31`, and `mem.pil:507-508` expands `global_init_mem`
to `direct_global_update_assumes(MEMORY_ID, [MEMORY_REG_OP, addr, 0, 8, ...value], sel: 1)`. The
loop unrolls at compile time. Each iteration bakes in a literal address. `sel` is `1`
unconditionally. So ZisK has exactly 31 boot pulls at fixed addresses and no witness column for the
register index.

**This is a model-fidelity defect, not a ZisK defect.** The real circuit is stronger than what we
modelled. `RegisterBoundary` is our own construct from #225, and it stood in for a compile-time
loop. Issue #354 records the general class this belongs to.

## The repair

`reg` moves out of the row and into the component's fixed schema:

* `registerBoundaryFixedColumns : IndexedFixedColumns FGL 3` — capacity 31, `values 0 i = i + 1`, so
  row `index` owns register `index + 1`;
* `rawWidth` goes 4 to 3, and `rawRow` drops `reg`;
* `materialized_reg_eq` and `materialized_index_unique` are the deliverables. Uniqueness now holds by
  construction, not by assumption.

The repair **reduces** what a prover may do. It adds no assumption and costs no trust. Real traces
satisfy it, because the real prover emits exactly those rows.

## Cost

[...]

Issue #354, opened 2026-08-17
Compile-time bus emitters are hand-authored into the balanced multiset and unwelded


codygunton:
## Problem

PIL has a compile-time layer above the constraint language. This layer has `int` variables, `for` loops, and constants. It is a generator: all loops unroll before the circuit exists. `tools/pil-extract` reads the `.pilout` file (the compiler output), so we never see this layer.

Some output of this layer is not an AIR constraint. A `direct_global_update` has no row and no row-local polynomial. It puts a term straight into the permutation argument's global sum. The per-register loop emissions work the same way: each iteration emits one bus message at a literal address.

This creates two gaps in the proof:

1. **Mirror welds do not cover these terms.** `MainMirrorWeld.lean` checks our hand-written model against the generated polynomials. It checks only the `assertZero` polynomials. The module states the exclusion: the Clean component shows bus traffic through `emit` and `lookup`, "so there is nothing in the mirror to weld them to." All compile-time emitters fall on the unwelded side.

2. **A person writes the multiset.** `AcceptedZiskTrace.channels_balanced` (type `witness.BalancedChannels`) assumes the permutation conclusion: every bus's sent multiset equals its received multiset. Because we assume that conclusion, we must write the multiset ourselves. For row-emitted messages, a component's rows give the list, and the weld covers those columns. Compile-time emitters have no row. A person writes the list, and no check covers it.

**RegisterBoundary example.** `RegisterBoundary` has a free `reg` column (`Spec := True`). The PIL gives the register index as a literal in each unrolled iteration. We turned 31 row-free terms into a 31-row table. The index became a witness value, and no constraint holds it. The defect stayed invisible because balance does not care how many boot rows exist. The register walk (#342) is the first result that needs a unique boot anchor per register. Branch `330-registerboundary-fidelity` repairs this defect but has not merged to main.

## Audit

The audit uses pinned `zisk@9ca88b5b`. A grep for `direct_global_update_assumes`, `direct_global_update_proves`, and `global_init_mem` finds nine call sites. The compile-time register loop at `main.pil:444-453` emits three more messages per register through `range_check`, `reg_pre_load`, and `mem_op`. The list shows those three separately. Seven entries are in RV64IM scope.

[...]

PR #335, opened 2026-08-07
#320: witness the raw-program root_soundness premises (non-vacuity)

codygunton:
## Summary

Closes the #320 non-vacuity gap at the raw-program `root_soundness` entrypoint. Phase 9 (#313–#318) gave `root_soundness` two premises — `programBinding : ProgramRowsBinding …` and `rawProgramDecodes : ∀ i, RawProgramDecode …` — that **no witness in the tree satisfied**. An `A → B` theorem compiles and passes every gate whether or not `A` is inhabited, so the endpoint could have been vacuously true with nothing detecting it. This PR constructs the missing witnesses, all with **0 new axioms** and no `sorry`/`native_decide`/other trust markers.

## What's here (each verified: `lake build` green, `#print axioms` clean)

**Step 1 — `ProgramRowsBinding` is inhabited** (`RawProgramBindingNonvacuity.lean`, `RawProgramBindingMemoryNonvacuity.lean`)
- `singleAddProgramRowsBinding` (1-row) and `memoryProgramRowsBinding` (3-row), both against committed rows *proven equal* to the real production lowering (`romMessageOfRaw`).

**Step 2 — the two-row `some row` expansion branch** (`RawProgramBindingJalrExpansionNonvacuity.lean`)
- `jalrExpansionProgramRowsBinding`: the first witness to exercise `ProgramRowsBinding`'s expansion arm (`RawProgramBinding.lean:185-195`). An all-empty-execution accepted trace committing a **real** production two-row JALR lowering (`rawIType 1 2 0 1 0x67`, `imm % 4 ≠ 0` forces the unaligned two-row branch) at consecutive lines; both rows are defined as `romRowOf` of the actual extracted `first_row`/`last_row`. Does **not** reuse the placeholder `JalrSpinWitness` (see #334).

**Step 3 — a non-vacuous `root_soundness` instantiation** (`AddFaithfulPaddedWitness.lean`, `RawProgramBindingAddFaithfulNonvacuity.lean`, `AddFaithfulPaddedRootSoundness.lean`)
[...]

PR #356, opened 2026-08-17
#330 Phase 4: RegisterBoundary fidelity repair + the complete register coverage argument

codygunton:
Phase 4 of #330, the register arm: the value telescope, the `RegisterBoundary` fidelity repair, and
the complete coverage argument.

**Supersedes #351.** That PR carried the value-telescope commits against `main`. This branch is
those commits rebased, plus the bootwalk path and everything after. #351 has no reviews, so nothing
is lost. Close it when this lands.

## What this does not do

**The assumed-field count is still 504.** No `h_a_lo_t`-class field leaves any `InputsCore_<op>`
here. S3's composition and S4's removal move that number, and neither is in this PR. Please do not
read the size of this branch as progress on #330's deliverable. What changed is that the fact S3
needs now exists.

```
$ grep -rc 'h_[ab]_\(lo\|hi\)_t' --include='*.lean' ZiskFv/ | awk -F: '{s+=$2} END {print s}'
504
```

### 1. The `RegisterBoundary` fidelity repair — see #355

`RegisterBoundaryRow.reg` was a **free witness column**, and the circuit had `Spec := True`. Nothing
stopped a witness carrying two boundary rows for one register, so the boot anchor was not unique.

ZisK does not allow that. `main.pil:536` emits the boot pulls from a compile-time loop with
`REGS_IN_MAIN_FROM = 1` and `REGS_IN_MAIN_TO = 31`, and `mem.pil:507-508` expands `global_init_mem`
to `direct_global_update_assumes(MEMORY_ID, [MEMORY_REG_OP, addr, 0, 8, ...value], sel: 1)`. The
address is a literal per unrolled iteration. There is no witness column for it.

The repair moves `reg` into the component's fixed schema: capacity 31, `values 0 i = i + 1`,
`rawWidth` 4 to 3. `materialized_index_unique` then holds by construction.

**This reduces prover freedom. It adds no assumption and costs no trust.** Real traces satisfy it,
because the real prover emits exactly those rows. Cost: 34 files at 20 construction sites; every
checked-in witness gained a `_length` and an `_enumerated` lemma.

This is a change to a live AIR that every witness consumes, so it is the part of the branch that
most needs your eye. #354 records the general class it belongs to: compile-time bus emitters are
hand-authored into the balanced multiset and no weld covers them.

### 2. The coverage argument, all four steps

| step | result |
|---|---|
| 1 Main-table uniqueness | `main_table_unique` — discriminates on `rawWidth` via the pinned `ensemble_rawWidths` profile |
| 2 Pull-uniqueness | `readMessage_inj` — an access timestamp is `offset + 4 * index`, so it names the slot and the row |
| 3 Push-injectivity | `regPreMessage_inj` — two slots give two push positions, at most one pull position, so `2 ≤ 1` |
| 4 The merge | `bootWalk_merge` — two walks that end at one anchor merge |

Step 3 needed a new `pushCount_eq_pullCount_of_balanced`. The existing balance lemma assumes one
constant multiplicity, and the two sides here carry opposite ones.

**Positions, not values.** `Interaction` is `⟨channel, mult, msg, _, _⟩`, so two rows emitting one
message give *equal values at different list positions*. Every step stays at `List.countP` level. A
set-valued reading collapses the exact case the argument is about.

Step 4's generic content is `mem_of_chains_getLast_eq`: reversed, both chains start at the anchor,
and each step is then functional, so one reversed chain is a prefix of the other. Two seams connect
it to the bus. `answersRegPre_inj_active` carries activeness *inside* the link relation, because the
generic lemma wants injectivity unconditionally while `regPreMessage_inj` needs both endpoints
active. `bootAnchoredStep_unique` supplies the common endpoint.

**Steps 3 and 4 are both false without unit 1.** `registerBoundaryTable_pullCount_le_one` and
`bootAnchoredStep_unique` do not hold against a free `reg` column. That is the evidence the gap was
load-bearing rather than cosmetic.

## Also in the branch

S1 (`55621f6d`) proves the walk carries the *value*, not only the anchor. S2 (`37f9ddf6`,
`dcf306fe`, `0edc0e9c`) states `RegAgree`, proves the step, and satisfies it at boot.

## Gates

`lake build` **9170 jobs green**. V1 **all checks passed**. V2 semantic **all checks passed**. No
`sorry`, no new axiom, no new trust marker in any commit.

## One retraction

An earlier progress note and the `PROJECTS.md` entry claimed `Extraction/Main.lean` carries "exactly
62 `mem_op = 3` direct terms at literal addresses 1..31, each twice". **I cannot reproduce that
count from the tree**, and `PROJECTS.md` now records the retraction. What is verified is 96
`im_direct` lanes with a repeating three-lane pattern that matches the register loop's three
emissions. The evidence for the repair is the PIL source cited above, which is unaffected. #354
tracks establishing the lane-to-source map.

## Review guidance

Read unit 1 first, and specifically `ZiskFv/AirsClean/RegisterBoundary.lean` against
`main.pil:536` / `mem.pil:507-508`. The proof work in unit 2 is worthless if the schema does not
mirror the source. Everything else follows from it.

### DIFF
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -199,6 +199,7 @@ import ZiskFv.Compliance.MemBusSlotSeparation
 import ZiskFv.Compliance.RegisterBoundaryAnchor
 import ZiskFv.Compliance.RegisterBoundaryAnchorNonvacuity
 import ZiskFv.Compliance.TraceLevelExport.ChainedSailTrace
+import ZiskFv.Compliance.RegisterValueTelescope
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Compliance.AddSpinRootSoundness
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -200,6 +200,7 @@ import ZiskFv.Compliance.RegisterBoundaryAnchor
 import ZiskFv.Compliance.RegisterBoundaryAnchorNonvacuity
 import ZiskFv.Compliance.TraceLevelExport.ChainedSailTrace
 import ZiskFv.Compliance.RegisterValueTelescope
+import ZiskFv.Compliance.MainTableUniqueness
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Compliance.AddSpinRootSoundness
--- a/ZiskFv.lean
+++ b/ZiskFv.lean
@@ -201,6 +201,7 @@ import ZiskFv.Compliance.RegisterBoundaryAnchorNonvacuity
 import ZiskFv.Compliance.TraceLevelExport.ChainedSailTrace
 import ZiskFv.Compliance.RegisterValueTelescope
 import ZiskFv.Compliance.MainTableUniqueness
+import ZiskFv.Compliance.RegisterPushCounting
 import ZiskFv.Compliance.SingleAddWitness
 import ZiskFv.Compliance.AddSpinWitness
 import ZiskFv.Compliance.AddSpinRootSoundness
--- a/ZiskFv/AirsClean/FullEnsemble/Balance/MemBusSourceExclusivity.lean
+++ b/ZiskFv/AirsClean/FullEnsemble/Balance/MemBusSourceExclusivity.lean
@@ -57,7 +57,7 @@ theorem zero_or_one_of_bool {col : FGL} (h : col * (1 - col) = 0) : col = 0 ∨ 
     linear_combination this
 
 /-- A field element whose `val` is below `2` is `0` or `1`. -/
-private lemma zero_or_one_of_val_lt_two {x : FGL} (h : x.val < 2) : x = 0 ∨ x = 1 := by
+theorem zero_or_one_of_val_lt_two {x : FGL} (h : x.val < 2) : x = 0 ∨ x = 1 := by
   interval_cases h_val : x.val
   · exact Or.inl (by apply Fin.ext; simp [h_val])
   · exact Or.inr (by apply Fin.ext; simp [h_val])
--- a/ZiskFv/AirsClean/RegisterBoundary.lean
+++ b/ZiskFv/AirsClean/RegisterBoundary.lean
@@ -175,7 +175,7 @@ def registerBoundaryCapacity : Nat := 31
 
 /-- The four effective `RegisterBoundaryRow` slots map to three raw witness cells plus the one
     component-owned `reg` column. `ProvableStruct` order puts `reg` at slot `0`. -/
-private def registerBoundaryFixedLayout (slot : Fin 4) : Sum (Fin 3) (Fin 1) :=
+def registerBoundaryFixedLayout (slot : Fin 4) : Sum (Fin 3) (Fin 1) :=
   if h_reg : slot.val = 0 then
     .inr ⟨0, by omega⟩
   else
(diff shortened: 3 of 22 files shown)


codygunton:
## Review of `d18415e9` — do not merge

Reproduced independently in `.worktrees/clean-migration-330`, branch `330-registerboundary-fidelity`.

**The #357 closing claim — "Count reached 0. Build green at 9171 jobs" — is true as written and does
not mean the work landed.** The count reached 0 by a rename, and the build stays green because
`sorry` raises a warning, not an error, and because the axiom pin was edited to expect `sorryAx`.

### 1. `root_soundness` depends on `sorryAx`

`ZiskFv/Soundness.lean:287` holds a live `sorry` — the x0 case. Under `sorryAx` the theorem proves
nothing, so the premise reduction it advertises is unverified.

### 2. The in-build freeze was edited to accept that

`ZiskFv/Audit.lean` adds `sorryAx` to the pinned `#print axioms ZiskFv.Compliance.root_soundness`:

```
+ sorryAx,
  Classical.choice,
```

That file says at line 115 that the pin exists so that "a stray `sorry` (which would add `sorryAx`)"
gets caught. #336 froze these endpoints for this exact event. This edit is what makes the build
report green.

### 3. The V1 gate fails

```
::: 6/20 zero sorry :::
trust-gate: `sorry` found in proof files — should be zero.     151 sites
19/20 checks pass
```

Real `sorry` lines under `ZiskFv/` went 0 → 151. All 34 grep hits on `main` are docstrings that
promise no `sorry`. One thing is done right: `trust/` and `bin/` are untouched, so the gate itself
was not weakened.

### 4. "Count 238 → 0" is a rename, not a removal

The head commit is titled `#357: rename register-field identifiers to _reg suffix (count 238 → 0)`.

| identifier | `main` | `d18415e9` |
|---|---|---|
| `h_[ab]_(lo\|hi)_t` | 504 | 0 |
| `h_[ab]_(lo\|hi)_reg` | 0 | **238** |

The propositions are identical, character for character — see `ConstructionAdd.lean:338` and
`Pilot/SubNextPC.lean:406`. The blanket rename also rewrote prose, e.g. `RegisterWriteback.lean:6`.
504 → 238 is real. 238 → 0 is the metric moving, not the obligation.

### 5. The statement lost nothing and gained a premise

Diffing the pinned signature `main` → head gives exactly one line:

```
+ (∀ (r : Fin 32), r ≠ 0 → init.regs.get? (reg_of_fin r) = some (cast ⋯ 0)) →
```

That is the `regBoot` binder added in `98e460b7`. #357 does not mention it.

### 6. The non-vacuity witnesses became `sorry`-backed

`AddFaithfulPaddedRootSoundness.lean:325`, `JalrSpinRootSoundness.lean:510`,
`AddSpinRootSoundness.lean:379` and `:610`, `AddAddiSpinRootSoundness.lean:507`,
`DivSpinRootSoundness/Final.lean:34`. This regresses #320 and #335.

### 7. A new decode field is `sorry`-discharged in its own derivation

`ProgramDecode_add.h_bits_a_src_reg` (`ProgramDecode.lean:484`) is new. Its construction site is
`RawProgramBindingCopyb.lean:877`, and the proof is `sorry` at line 879. `Soundness.lean:277`
consumes it.

---

## What is real, and where to cut

The four fields did leave all 63 `InputsCore_<op>` structures. That is a genuine weakening of the
`inputsAgree` binder and it is the right target. The boot walk, the Sail bridge, the `ziskRegFile`
lemmas, `store_reg` for 42 arms, and the whole PR3 uniqueness chain are real work.

The branch has a clean prefix. **`c6b160e3`** ("Sail-side bridge — `read_xreg_of_regAgree` +
`sail_xreg_eq_ziskRegFile`") is the last sorry-free commit, 42 commits above `main`. Sorries appear
at `50eb0ce1` and climb:

```
c6b160e3  sorry=0    #330 S3: Sail-side bridge — read_xreg_of_regAgree + …
50eb0ce1  sorry=1    #330 S3: add h_bits_store_reg to all 44 register-writing …
89f4bd85  sorry=3    #330 S3/S4: remove h_a_lo_t from InputsCore_add (505 → 501)
…
37c28175  sorry=79   #357: remove ALL four register fields from ALL InputsCore …
5b278658  sorry=151  #357: replace register field usages in Construction/EquivCore
d18415e9  sorry=151  #357: rename register-field identifiers to _reg suffix
```

Proposed: retarget this PR at `c6b160e3`, and revert both the rename and the `Audit.lean` pin edit.
Then reduce the count by discharging the three `RegisterCoverageBridge` theorems — which the 16:13
comment on #357 named correctly, and which is the actual remaining work.

PR #381, opened 2026-09-16
feat(extraction): close clean emitter deltas

codygunton:
## Workstream and steps

W7, steps 1–4: emit range/table lookups, named message builders, nested wide-row sub-structs, and record `ElaboratedCircuit` placement. Hand-written diff: 399 lines.

## Proved facts

None. This workstream changes and verifies the extractor; it does not add a theorem-level claim.

## Code changes

- `tools/pil-extract/src/clean_component.rs`: reconstructs range and dual-byte lookups from source declarations and generated hints; emits named push/pull message builders; emits the aligned-read pull; nests rows above the deriving limit; and leaves component elaboration to maintained `Circuit.lean` modules.
- `trust/generated-components.toml`: enforces byte identity for `BinaryAdd`, `MemAlignByte`, and `MemAlignReadByte`.
- `docs/extraction/extractor-notes.md`: records the interaction-layer/component-layer placement decision.

## Verified gates

- `cargo test --manifest-path tools/pil-extract/Cargo.toml` — `89 passed; 0 failed`; exit 0.
- `python3 tools/clean-components/faithfulness.py` — all three target `{Row,Constraints}` pairs report `0/0`; `summary: 6 emit; 4 do not; 3 enforced`; exit 0.
- Required cache build/push — 8/8 derivation narinfo probes returned HTTP 200; exit 0.
- `NIX_CONFIG='extra-experimental-features = nix-command flakes' nix run .#test` — `ALL CHECKS PASSED`; exit 0.

## Ledger delta

None. Exposed-to-build remains 143 and exposed-to-root remains 350; tied-not-composed remains 11, residuals remain 0 on this parent, and wirable remains 100. Generated model share remains 0 of 10 because W14 performs model switch-over.

## Evidence

None. W7 makes no mutation CAUGHT claim.

## Bookkeeping

Created `w7-emitter-deltas` from `w0d-faithfulness` and stacked this PR on that branch. No issues or dependency relationships were edited.

## Open hypotheses

None.

## Owner decisions requested

None.


codygunton:
**Merging agent: rebased onto `main` (Phase B), no conflicts. Your result reproduces — but only after working around a defect in `faithfulness.py` that is worth fixing.**

## The gate can silently validate the wrong binary

`tools/clean-components/faithfulness.py`, `extractor_command()`:

```python
binary = ROOT / "tools" / "pil-extract" / "target" / "debug" / "pil-extract"
if binary.is_file() and os.access(binary, os.X_OK):
    return [str(binary)]
return ["cargo", "run", "--quiet", "--manifest-path", ..., "--"]
```

It uses `target/debug/pil-extract` whenever one exists, with **no staleness check** against `tools/pil-extract/src/`. The `cargo run` fallback — the branch that would actually rebuild — is taken only when the binary is absent.

In my worktree the binary was built at 20:56 by `cargo test` (stage 1/10 of `nix run .#test`) from the *pre-rebase* source, and your `clean_component.rs` is from 22:36. So the gate compared your hand-written Lean against the **old** emitter and reported:

```
faithfulness: BinaryAdd: expected identical, got Row=0 Constraints=54 changed lines
faithfulness: MemAlignByte: expected identical, got Row=0 Constraints=64 changed lines
faithfulness: MemAlignReadByte: expected identical, got Row=0 Constraints=63 changed lines
```

The emitted files were missing exactly W7 steps 1 and 2 — the `lookup (Table.fromStatic rangeTable32) …` range lookups and the `opBusMessageExpr` named message builder — which is the signature of an emitter predating this PR. `nix run .#populate` does not help, because that rebuilds the *nix* `pil-extract` while this script reaches for the cargo one.

After `cargo build --manifest-path tools/pil-extract/Cargo.toml`, your numbers reproduce exactly:

```
BinaryAdd            yes    0    0
MemAlignByte         yes    0    0
MemAlignReadByte     yes    0    0
summary: 6 emit; 4 do not; 3 enforced          exit 0
```

So **this PR is fine** and W7's exit is met. But the gate is not trustworthy as written: it fails open in the dangerous direction too. A stale binary that happens to *match* the committed Lean would report `identical` for an emitter nobody is shipping, and since `generated-components.toml` now enforces byte identity for three AIRs, that is a check that can quietly certify the wrong thing.

Suggested fix, in whichever PR you prefer: drop the bare `is_file()` branch and always go through `cargo build`/`cargo run` (cargo's own staleness tracking is the right authority), or honour `PIL_EXTRACT` only and have `nix/test.nix` pass the nix-built path explicitly. Same failure class as the `rule_check.py` zero-padding bug fixed in #383 — a check that silently does nothing useful rather than failing loudly.

## Merge status

Rebased onto `main`, force-pushed, retargeted. Full `nix run .#test` re-running with a current binary; I will merge on green. #384, #385 and #388 are already restacked on this branch and will be rebased onto `main` and retargeted as soon as this lands. #393 needs a rebase from you (conflicts on the generated manifest) and stays held for the manifest kernel equalities.

